PR #49, opened 2014-09-11
Filter colorschemes

jubalh:
Similar to vim, only display colorschemes with file extension '.vifm'.

Idea arose through through a [pull request in vifm-colors](https://github.com/vifm/vifm-colors/pull/2).

### DIFF
--- a/src/color_scheme.c
+++ b/src/color_scheme.c
@@ -38,6 +38,8 @@
 #include "status.h"
 #include "ui.h"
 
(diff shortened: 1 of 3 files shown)


xaizek:
I thought about it a bit more and extensions should be introduced sooner or later, so did this:

> 1. If there is at least one file with .vifm extension, filter out all files without such extension.
> 2. Otherwise, list all files.

Plus automatic appending of extensions on first run (when there is no files which names end with `.vifm`).

Thanks.

PR #86, opened 2015-04-02
Updating license file

jubalh:
The COPYING file states an old version of the GPLv2.
I updated the file with the latest version.

The legal part itself do not change. It's just about updating the adress
of the FSF and some things like "19yy" to "year".
However there seem to be some nonvisual character changes in that file
too. Nevertheless I it is the offical version available at
https://www.gnu.org/licenses/gpl-2.0.txt that I added here.


xaizek:
Thanks!

PR #107, opened 2015-08-07
Added * for slowfs

cosminadrianpopescu:
I've added the possibility to set the `slowfs` option to `"*"`. This means that all directories are treated as slow. I did this because on `cygwin`, calling the `traverse_mount_points` and `start_with_list_item` function every second is very slow. 

So even if I am on a slow file system, and the system is included in the `slowfs` option, the overall experience is extremely poor. Setting the `slowfs` to `"*"` makes all the difference. `vifm` becomes usable also under `cygwin`.

There is also the possibility to check at a compile time with the preprocessor that `vifm` is running in `cygwin` and disable the `slowfs`, but maybe there are people who want to use it even slow under `cygwin`. So to me, this seems like the better option.

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -3312,7 +3312,9 @@ you.  This option can be used to stop vifm from making some requests to
 particular kinds of file systems that can slow down file browsing.
 Currently this means don't check if directory has changed, skip check if
 target of symbolic links exists, assume that link target located on slow fs
-to be a directory (allows entering directories and navigating to files via gf).
+to be a directory (allows entering directories and navigating to files via gf). 
+If you set the option to "*", it means all the systems are considered slow
+(usefull for cygwin, where all the checks will render vifm very slow). 
 
 Example for autofs root /mnt/autofs:
 .EX
--- a/data/vim/doc/app/vifm-app.txt
+++ b/data/vim/doc/app/vifm-app.txt
@@ -2767,7 +2767,9 @@ particular kinds of file systems that can slow down file browsing.
 Currently this means don't check if directory has changed, skip check if
 target of symbolic links exists, assume that link target located on slow fs
 to be a directory (allows entering directories and navigating to files via
-|vifm-gf|).
+|vifm-gf|). If you set the option to "*", it means all the systems are
+considered slow (usefull for cygwin, where all the checks will render vifm
+very slow). 
 
 Example for autofs root /mnt/autofs: >
   set slowfs+=/mnt/autofs
--- a/src/utils/utils_nix.c
+++ b/src/utils/utils_nix.c
@@ -384,6 +384,15 @@ is_on_slow_fs(const char full_path[])
 		.curr_len = 0UL,
 	};
 
+	/* if slowfs = "*" then all file systems are considered slow
+	 * this function is very slow on cygwin
+	 */
+
+	if (strcmp(cfg.slow_fs_list, "*") == 0)
+	{
+		return 1;
+	}
+
 	/* Empty list optimization. */
 	if(cfg.slow_fs_list[0] == '\0')
 	{


xaizek:
Such slowdown is quite surprising and I can't reproduce it. How many entries are there in your `/etc/mtab` file and are there any references to network/removable devices?

I like the pull request, just want to understand what's the actual cause of the slowdown. Given that `/etc/mtab` is not reparsed unless it's changed and string comparisons cost basically nothing, real reason of performance degradation is not that clear.

cosminadrianpopescu:
Quite a few lines in `/etc/mtab` and most of them are network share. I will give you an exact number on Monday, when I go again to work, but I think there are more than 10 network shares in `/etc/mtab`

xaizek:
Thanks, then it might be the cause. I'm not sure what invalidates cache though, so will do some tests. Maybe this can be solved differently. The thing is that treating all file system types as "slow" disables checks for directory changes, which is undesirable, unless we don't really have choice.

cosminadrianpopescu:
I know that it disables the checks for directory changes, but this for me is not such a big deal. How can I check if the cache is invalidated? Where is the caching stored?

xaizek:
It's [at the top of `traverse_mount_points()`](https://github.com/vifm/vifm/blob/master/src/utils/utils_nix.c#L453-L468) and it doesn't work for cygwin. `/etc/mtab` is a symbolic link to `/proc/mounts` which links to `/proc/self/mounts`. Modification time for those pseudo-files is equal to the time of accessing them, so they are always "modified".

I guess this implies that `cat /etc/mtab` is also slow on your configuration and we really have no choice but skip reading it (we could calculate checksum if reading was cheap).

cosminadrianpopescu:
Thank you for testing it. So are you going to include the patch in vifm? As I was saying, from my point of view, I don't have a problem with the fact that the folder is not refreshed. I have `nnoremap <c-r> cd .<cr>` and I do `C-r` when I want to refresh it. I also don't use the `redo` feature. Anyway, anybody can map anything to do a manual refresh. 

Unrelated with this issue, I would be interested to contribute to the development of vifm. What do I need to do?

xaizek:
> So are you going to include the patch in vifm?

Already did, it just wasn't pushed. Thank you!

> I would be interested to contribute to the development of vifm. What do I need to do?

"contribute" is ambiguous word. I'd assume you mean contributing changes, but you know how to do that. So in case you are talking about donations, there is corresponding folded section on the main page of [the site](http://vifm.info).

cosminadrianpopescu:
No, sorry, I wasn't clear. I would like to help with the development if such help is needed. I am a developer myself, I use `vifm` as my file manager and I love C. I've had a quick look in the `TODO` list and I've seen a lot of things that I would like to implement from there. 

I have the technical knowledge. Another thing that I would like (I really miss it in `vifm` and is not on `TODO` list) are for example tabs. Just like a `tabnew` in `vim`. Before I was using `mc` and I've added tabs myself (see https://www.midnight-commander.org/ticket/1581).

xaizek:
> I would like to help with the development if such help is needed.

That would be great! I could use some help to improve vifm faster.

> I love C

I love hearing that in this age of "web-everything".

> I've had a quick look in the TODO list and I've seen a lot of things that I would like to implement from there.

It's not complete list actually, but items there are relevant.

> Another thing that I would like (I really miss it in vifm and is not on TODO list) are for example tabs.

I'm missing them too in vifm and was going to try to add tabs for the next release, but I didn't start yet. So go for it if you wish.

In general, pick whatever you like, just let me know so that I'm aware (and won't do the same) and could give you some advise on implementation (and share relevant instructions I have on adding new files/commands/options/etc., which I'll upload to the wiki then). You can reach me here, on irc (when I don't forget to start it...) or via email in case of questions. The code and/or its organization is not always obvious, so feel free to ask questions.

cosminadrianpopescu:
So, I was thinking of adding tabs. In the beginning, I would add `tabnew`, `tabclose`, `tabnext`, `tabprev`, `tabfirst`, `tablast` and the `tabs` commands. 

The tabs would be per panel. So each panel would have its own tabs (like any other twin panel file manager with tabs). 

From a technical point of view, I would have a structure with two fields: `name1 and `path` for each tab. If the name is `NULL`, then the name would be the current folder. The `tabnew` command would take an optional argument, which would be the name of the tab. 

I would display the tabs where the the path is displayed. 

When doing a `tabnext` or `tabprev`, for example, I would simply do a `cd` to the path of the tab in which I am suppose to land. 

Please let me know what do you think. Also, are there any code guidelines to follow?

Thank you, 
Cosmin Popescu.

xaizek:
> The tabs would be per panel. So each panel would have its own tabs (like any other twin panel file manager with tabs).

I imagined it to be like tabs in Vim, where tab contains windows (here panes) and as it stands, preview uses both windows and you can't separate them easily.  Say you do `:view` and then switch tab on a single pane, both of them are affected.
At least this thing with preview should be addressed somehow.

> From a technical point of view, I would have a structure with two fields: name and path for each tab. If the name is NULL, then the name would be the current folder.
> When doing a tabnext or tabprev, for example, I would simply do a cd to the path of the tab in which I am suppose to land.

I'm afraid it won't work.  Each pane contains quite a lot of data including selected items, items themselves for custom views (you can't just drop them, because there is no way to restore them), state of filters.
When switching between tabs they shouldn't be reset each time, it's not very useful when working on multiple things simultaneously.
There is additional complication related to history, which should be managed in predictable way (e.g. be local to pane-tab pair), although this doesn't need to be considered right from the start.

I was thinking about storing [FileView](https://github.com/vifm/vifm/blob/master/src/ui/ui.h#L166) structures and backing up/restoring them on switching tabs.  This should mostly work as is except that some fields might need to be shared among all tabs or cloned on tab creation.

> The tabnew command would take an optional argument, which would be the name of the tab.

I like this.

> I would display the tabs where the path is displayed.

Will it be enough space to include whole path?  This is quite important thing for file manager to display current location, and on many tabs there won't be much place.
One thing that bothers me in Vim is that there is no way to see all tabs on very long tabline, something should be done to omit this unfortunate behaviour and ensure that at least current tab and those around it are always visible.

> Also, are there any code guidelines to follow?

Yes, please keep it close to existing code, which has almost consistent style.
I put an old template on the wiki (started some time ago, but didn't finish), which is [here](http://wiki.vifm.info/index.php?title=Code_Guidelines).  I'll try to put more details there.

Hope I didn't scare you :)  Maybe consider doing something simpler to get used to the code, adding tabs has some difficulties.

cosminadrianpopescu:
> I imagined it to be like tabs in Vim, where tab contains windows (here panes) and as it stands, preview uses both windows and you can't separate them easily. Say you do :view and then switch tab on a single pane, both of them are affected.
> At least this thing with preview should be addressed somehow.

The reason I don't really like this way of handling tabs is because I can very easily achieve this using `tmux`. So I don't think it's worth the effort. When I added tabs to `mc` I added them per panel from the very same reason. Right now (at work) I use `tmux` and basically I have this kind of tabs. 

> Will it be enough space to include whole path? This is quite important thing for file manager to display current location, and on many tabs there won't be much place.
> One thing that bothers me in Vim is that there is no way to see all tabs on very long tabline, something should be done to omit this unfortunate behaviour and ensure that at least current tab and those around it are always visible.

This is true, but I was thinking of having a `<` or `>` characters to suggest scrolling (left or right).

> Hope I didn't scare you :) Maybe consider doing something simpler to get used to the code, adding tabs has some difficulties.

No, you didn't scare me. Don't worry. Could you then please suggest something simpler for me to do? Please note that I don't have a Windows at home, so I can only work on Linux features and then (if they are suppose to also work on the windows version) I can test them at work.

xaizek:
> The reason I don't really like this way of handling tabs is because I can very easily achieve this using tmux. So I don't think it's worth the effort. When I added tabs to mc I added them per panel from the very same reason. Right now (at work) I use tmux and basically I have this kind of tabs.

The good things is that per pane tabs are flexible enough to implement regular tabs, should such need arise (say with option like `'bindtabs'`). So it's not an issue.

> This is true, but I was thinking of having a < or > characters to suggest scrolling (left or right).

Vim has such marks, but no way of scrolling, which is frustrating sometimes. I was just saying that we should avoid such situation.

> Could you then please suggest something simpler for me to do?

You said you would like to implement some items from TODO list, which ones?

One more item, which isn't on the list: displaying match number for search (e.g. `(2 of 4) /pattern` instead of current `/pattern`), the information is basically there, but requires changing a bit, accounting for sorting and displaying.

> Please note that I don't have a Windows at home, so I can only work on Linux features and then (if they are suppose to also work on the windows version) I can test them at work.

That's fine, I also use Windows only when I have to. Most of the time it's enough to try it out periodically and fix accumulated issues, which are usually related to compilation. I believe no one builds vifm for Windows himself (it's not that easy), so temporary breakage there is not a problem.

cosminadrianpopescu:
Ok, I think I will do the search thing from pattern for start. So, display `(2 of 4) /pattern`. This sounds great. As methodology of work, I suppose I can clone your repository, make the changes (every day make a `git pull`) and then send you the patch. Is this how you work? If not, could you please let me know how. 

Thank you, 
Cosmin Popescu.

xaizek:
> Ok, I think I will do the search thing from pattern for start.

Great. To save you some time: there is a [boolean field](https://github.com/vifm/vifm/blob/master/src/ui/ui.h#L154) which I'd re-purpose for holding match number, which can be updated in [sort_view](https://github.com/vifm/vifm/blob/master/src/sort.c#L62). I didn't try it, so not sure if that's enough to make it work.

> As methodology of work, I suppose I can clone your repository, make the changes (every day make a git pull) and then send you the patch. Is this how you work?

Pretty much it, although you might not need to rebase often, just before the pull request to keep the history more straight. Or if there is complicated conflict, send as is and I'll resolve it. And you don't need to clone this repository if you have your fork, just push branches there and create pull requests as in regular workflow on GitHub/GitLab and alike.

One quirk about git commit messages (maybe I need to add `CONTRIBUTING.md`), please follow guidelines from [this post](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). I find recommendations regarding formatting/language presented there quite reasonable.

> Thank you

Thank you!

PR #108, opened 2015-08-09
Update mntent.c

grassofhust:
typo.


xaizek:
Thanks! Didn't build on BSDs after that change.

Issue #54, opened 2014-10-12
viewcolumns: ext column and dotted files


ranousse:
With viewcolumns the extension of the file `.vimrc` is `vimrc`.
I would expect the extension of such a dotted file to be nothing.
I would also find more natural that directories always have no extension.

PR #109, opened 2015-08-09
Displaying match number for searches

cosminadrianpopescu:
On a search result, display `(1 of 3) /pattern`. 

This is implemented for searches in normal mode and in visual mode and for searches in menus.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -476,7 +476,7 @@ static const cmd_add_t commands[] = {
 		.handler = screen_cmd,      .qmark = 1,      .expand = 0, .cust_sep = 0,         .min_args = 0, .max_args = 0,       .select = 0, },
 	{ .name = "set",              .abbr = "se",    .emark = 0,  .id = COM_SET,         .range = 0,    .bg = 0, .quote = 0, .regexp = 0,
 		.handler = set_cmd,         .qmark = 0,      .expand = 0, .cust_sep = 0,         .min_args = 0, .max_args = NOT_DEF, .select = 0, },
-	{ .name = "shell",            .abbr = "sh",    .emark = 0,  .id = -1,              .range = 0,    .bg = 0, .quote = 0, .regexp = 0,
+	{ .name = "shell",            .abbr = "sh",    .emark = 1,  .id = -1,              .range = 0,    .bg = 0, .quote = 0, .regexp = 0,
 		.handler = shell_cmd,       .qmark = 0,      .expand = 0, .cust_sep = 0,         .min_args = 0, .max_args = 0,       .select = 0, },
 	{ .name = "sort",             .abbr = "sor",   .emark = 0,  .id = -1,              .range = 0,    .bg = 0, .quote = 0, .regexp = 0,
 		.handler = sort_cmd,        .qmark = 0,      .expand = 0, .cust_sep = 0,         .min_args = 0, .max_args = 0,       .select = 0, },
@@ -3780,7 +3780,7 @@ shell_cmd(const cmd_info_t *cmd_info)
 
 	/* Run shell with clean PATH environment variable. */
 	load_clean_path_env();
-	shellout(sh, 0, 1);
+	shellout(sh, 0, cmd_info->emark ? 0 : 1);
 	load_real_path_env();
 
 	return 0;
--- a/src/modes/menu.c
+++ b/src/modes/menu.c
@@ -750,7 +750,8 @@ search(int backward)
 
 		if(menu->matching_entries > 0)
 		{
-			status_bar_messagef("(%d of %d) %c%s", get_match_index(menu), menu->matching_entries,
+			status_bar_messagef("(%d of %d) %c%s", get_match_index(menu),
+					menu->matching_entries,
 					backward ? '?' : '/', menu->regexp);
 		}
 	}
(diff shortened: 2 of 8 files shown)


cosminadrianpopescu:
So, I wanted the third commit to be another pull request. I don't know why github included in this one (probably I did something wrong). 

However, I added an emark to the shell command. This is to not consider the tmux integration. I did this because I love the tmux integration, but not for shells. I love viewing another file in another tmux panel, but the shell I wanted always in the same window with my manager. The reason for this is that is very handy to go to shell, go a `grep`, exit from shell, check something and do again a shell and have the `grep` results still displayed. 

If I do a `grep` with the tmux integration, the moment I do exit, I am going to lose my `grep` results. The solution is not to exit, just to change the `tmux` window. But the problem is that if then I do another shell in another folder, I have another tmux window. And I don't know every time when I exit shell if I will need the results anymore or not. 

So with the e-mark for the shell, I can have the tmux integration active, and then map the `s` key to `:shell!`. Like this, I have the files which I want to view or edit opened in different `tmux` windows and the shell open in the same window. 

Please let me know if this is clear. If it's clear, please let me know what you think. 

Thank you, 
Cosmin Popescu

xaizek:
> So, I wanted the third commit to be another pull request. I don't know why github included it in this one (probably I did something wrong).

Pull requests are bound to branches (`master` in this case) and track their changes. To put commits in separate pull requests, you need to create dedicated branches for those commits.

> However, I added an emark to the shell command.

That's nice, I did `:$SHELL%n` in such cases, which is way less convenient than simple `:shell!`.
By the way, you could also use `:grep` in vifm when you need to open multiple files from `grep` results list.

The changes look good except that item numbers can be reordered after resorting, that's why I mentioned sorting procedure, where numbering can be restored. Try this:
1. Search for something to get multiple matches.
2. Resort with, say, `:set sort=-atime`.
3. Walk through matches, which will be out of order (e.g. 3, 5, 1, ...).
Please address this if you can or I can add it during merge.

Thanks, I find messages after these changes way more informative than before.

cosminadrianpopescu:
> That's nice, I did :$SHELL%n in such cases, which is way less convenient than simple `:shell!`.
I didn't knew about `%n`. I've just read now the docs. 

> By the way, you could also use :grep in vifm when you need to open multiple files from grep results list.
That I knew, but it's not just about the `grep`. I was just giving an example. As a rule, I prefer to have all the operations that will get away the explorer in a separate window, with the exception of the shell. 

> Please address this if you can or I can add it during merge.
I will fix this tomorrow and submit it.

cosminadrianpopescu:
Just fixed the bugs. There was also a problem with the search in menus.

cosminadrianpopescu:
I was looking at the `TODO` list for what to do next, but from the basic things, only those are clear to me: 

`Search match highlight in menus.`
`For 'syscalls', vifm should ask for standard retry/abort/cancel actions on error.`

I suppose I can make one of these. For others I don't understand 100%. So, is it ok if I do this 2 things? Do you have other things more important / urgent to be done? 

Thank you, 
Cosmin Popescu.

xaizek:
> Just fixed the bugs. There was also a problem with the search in menus.

Thanks  for the update, I'll merge the changes soon.

> I was looking at the TODO list for what to do next, but from the basic things

Mind that "basic" doesn't mean "easy" here, it's more like fundamental/important.

> For 'syscalls', vifm should ask for standard retry/abort/cancel actions on error.

This one is quite big and complicated actually, there is a lot of operations that potentially need to be changed to handle this properly.

> Search match highlight in menus.

This one should be easier, you probably seen [`match*` fields](https://github.com/vifm/vifm/blob/master/src/ui/ui.h#L155-L156), which are set in [`find_pattern()`](https://github.com/vifm/vifm/blob/master/src/search.c#L150-L151). Note that highlighting itself might be tricky, see [`highlight_search()` for views](https://github.com/vifm/vifm/blob/master/src/fileview.c#L884), although in menus there is only left alignment (which removes half of that function). Also, most of `menus.c` was written long time ago and it seen no proper update since then, code that draws menu is a mess, so be prepared if you will take this item.

cosminadrianpopescu:
> Mind that "basic" doesn't mean "easy" here, it's more like fundamental/important.

That's what I was thinking also. 

Then I will work on the highlighting on the menus.

cosminadrianpopescu:
I was also thinking at a new sorting option. I would like for the extension sorting, to have an option to sort only the files by extension. Folder would continue to be sorted by the name, not by extension. Usually the extension is not relevant for a folder. In most cases the extension is relevant only for files. For folders it means just a version number. 

What do you think?

xaizek:
I remember @ranousse mentioning something similar in #54, but I didn't want to make this the only behaviour and didn't think about adding new sorting key (something like `fileext` if you didn't think of a name yet). It might be worth providing this as an additional variant, I agree it makes sense in quite a lot of contexts.

cosminadrianpopescu:
So is it ok if I also implement this? I will make the highlighting of the menu for searching and also this if you agree.

xaizek:
> So is it ok if I also implement this? I will make the highlighting of the menu for searching and also this if you agree.

Yes, sure. I meant that it seems to be useful for at least several people, so deserves to be configurable. Sorry if my reply wasn't clear.

PR #111, opened 2015-08-12
Added fileext option

cosminadrianpopescu:
* Added an option `fileext` sort type. The files are sorted by extension, the folders by name 
* Added a new column in the viewcolumns: `fileext`. The column will contain the extension for files, nothing for folders.
* Fixed a small inconsistency: If sorting by extension, the hidden files (`.name`) will be sorted by their name, ignoring the point, but considered without extension. But when displaying the `ext` column in the `viewcolumns`, that would be empty for hidden files, which is correct. The fix is that when sorting, the hidden files will be considered as files without extension and will appear on top of all other files, under the folders. Please let me know what you thing about this.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -94,4 +94,5 @@ in command-line mode.
 
 Brian Callahan (a.k.a. ibara) fixed build on OpenBSD.
 
-Cosmin Popescu (a.k.a. cosminadrianpopescu) added "*" to 'slowfs'.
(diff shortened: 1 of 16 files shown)


xaizek:
Thank you, I'll pull the changes in tomorrow.

> The fix is that when sorting, the hidden files will be considered as files without extension and will appear on top of all other files, under the folders. Please let me know what you thing about this.

Didn't notice this issue before, thanks for fixing it.

PR #113, opened 2015-08-22
new align option

cosminadrianpopescu:
Added the AT_FILE alignment option. This is just as left, but when the text exceeds the boundaries, then the alignment will be switch to right. Like this, the file name will always be visible. The name is not very inspired, maybe you have a better idea. 

Also fixed the logging option. The /var/log/vifm-startup-log was hard coded for the log location, and because of the permissions, it was not created. Now the log file is saved in ~/.vifm/vifm-startup-log.


xaizek:
> Added the AT_FILE alignment option. This is just as left, but when the text exceeds the boundaries, then the alignment will be switch to right. Like this, the file name will always be visible. The name is not very inspired, maybe you have a better idea.

Thanks, this might be useful, especially in custom views.

> Also fixed the logging option. The /var/log/vifm-startup-log was hard coded for the log location, and because of the permissions, it was not created. Now the log file is saved in ~/.vifm/vifm-startup-log.

We can't use `~/.vifm` at that point because actual location of configuration is unknown yet, otherwise we wouldn't need separate startup-log file and could use regular one. I'll probably change that to location under temporary directory concatenated with user name. I think I saw this issue, but wasn't sure such rights are on every distribution, thanks for bringing it up.

cosminadrianpopescu:
> We can't use ~/.vifm at that point because actual location of configuration is unknown yet, otherwise we wouldn't need separate startup-log file and could use regular one. I'll probably change that to location under temporary directory concatenated with user name. I think I saw this issue, but wasn't sure such rights are on every distribution, thanks for bringing it up.

Then I can implement something like --logging=<path-to-log>

xaizek:
> Then I can implement something like --logging=<path-to-log>

Hm, that's a valid option.  Maybe don't log anything at startup by default (it's useless in most cases anyway) and do log the rest to `<config>/log`, but if `<path-to-log>` is provided log everything to that location?

cosminadrianpopescu:
> Hm, that's a valid option. Maybe don't log anything at startup by default (it's useless in most cases anyway) and do log the rest to <config>/log, but if <path-to-log> is provided log everything to that location?

Ok. I will implement something like this.

xaizek:
I've merged in align part. Renamed it to "dynamic" as real alignment is chosen dynamically.

Thanks!

Issue #103, opened 2015-07-20
Use of MAX_ARG_STRLEN breaks compilation on AArch64


hrw:
MAX_ARG_STRLEN definition depends on PAGE_SIZE to be defined. But on AArch64 page size can be either 4K or 64K so it is not defined in any system header.

As a result vifm 0.8 fails to build on this architecture.

Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1244855

xaizek:
Thanks, added test to `configure` that actually checks whether `MAX_ARG_STRLEN` is usable. This looks like an oversight in kernel headers, providing macro that can't be used is wrong, it's either shouldn't be defined at all or should have a value that won't fail during compilation.

PR #115, opened 2015-08-28
dynamic page size for powerpc

jubalh:
utils-nix.c needs dynamic PAGE_SIZE for powerpc
as not defined by default.

We received this patch into openSUSE from:
Michel Normand <normand@linux.vnet.ibm.com>

### DIFF
--- a/src/utils/utils_nix.c
+++ b/src/utils/utils_nix.c
@@ -263,6 +263,9 @@ char **
 make_execv_array(char shell[], char cmd[])
 {
 #ifdef HAVE_MAX_ARG_STRLEN
+#ifndef PAGE_SIZE
+#define PAGE_SIZE sysconf(_SC_PAGESIZE)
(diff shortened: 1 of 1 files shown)


jubalh:
Hello xaizek,
I tried to find out if this is a valid addition. But please check it yourself, too.

xaizek:
Hi Michael,

this was solved in a different way for AArch64 (see #103) by checking `HAVE_MAX_ARG_STRLEN` to be usable in `configure`. I personally think that kernel headers are wrong here, they shouldn't define unusable macros (and assuming dependency on `PAGE_SIZE` like here isn't future proof). Yet this workaround makes sense if accompanied with check in `configure`, so I'll combine two approaches. Thanks!

xaizek:
The commit became d3503c796f526e8475bdc153 after rebase.

PR #170, opened 2016-03-02
Dialog shortcuts

oo-:
I added shortcuts for dialog options, because I often toggle between sort orders.

Thus with the default mapping of **S** to open the sort dialog:
* **Ss** sorts pane items by size
* **Se** sorts pane items by extension
* **SN** sorts pane items by name, ignoring case
* And so on...

dialog-shortcuts

For completeness I also added [shortcuts to the change dialog](https://github.com/oo-/vifm/blob/76fe047f9880d618710cf6a375a43bdee6042c25/src/modes/dialogs/change_dialog.c#L134). Shortcuts did not seem relevant for msg or attr dialogs.

I tried to preserve the style of existing code and UI. I did not add any tests because I didn't understand vifm's testing scheme.


xaizek:
Thanks! Wanted to do something similar for a while, but couldn't pick a way to label elements: numbers are inconvenient to type and letters might conflict with input in the future, but letters seem to be the most convenient to type.

PR #188, opened 2016-05-09
Enabled IRC notification

onovy:


xaizek:
Wouldn't it be annoying and is it of interest to anyone on the channel if things break for a while? Such notifications seem to be more appropriate for dev mailing list or something similar.

onovy:
you are right, so closing as won't fix :)

PR #201, opened 2016-08-24
Open all selected files with sxiv

Nicop06:
sxiv is efficient when handling many images. Take advantage of this when selecting multiple image files in vifm.


xaizek:
Makes sense, thanks!

PR #219, opened 2016-10-20
Make mime type detection a bit safer

sizeofvoid:
- use fork() to call magic_\* functions
- call file(1) in a child process

### DIFF
--- a/src/int/file_magic.c
+++ b/src/int/file_magic.c
@@ -31,6 +31,7 @@
 #include <stddef.h> /* size_t */
 #include <stdlib.h> /* free() */
 #include <stdio.h> /* popen() */
+#include <unistd.h> /* read() */
 
 #include "../utils/path.h"
 #include "../utils/str.h"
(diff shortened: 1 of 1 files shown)


xaizek:
Could you please explain the motivation? Especially for replacement of `popen()` with manual forking, while `popen()` does pretty much the same internally, but also wraps it in a `FILE` stream. Running `magic_*` functions in another process could protect from crashes in libraries, but that seems relatively rare event.

sizeofvoid:
The motivation was simple: isolate `magic_*` and file(1)/execl(3) calls from vifm. In the OpenBSD manpage from [popen(3)](http://man.openbsd.org/popen.3) you can find under section **CAVEATS** my reason to replace this call.

xaizek:
The man page says:
> Never supply the popen() function with a command containing any part of an unsanitized user-supplied string. Shell meta-characters present will be honored by the sh(1) command interpreter.

But file name *is* sanitized, special characters are escaped by `shell_like_escape()`, which is used in many other contexts and if we trust it there, I don't see why one specific use case should differ. This also hard-codes path to `file`, while somebody might want to have it in different place (might be worked around by using `/usr/bin/env` though).

I don't really like overhead of forking to get mime-type, but since it's already too slow to be used for file highlighting, the difference shouldn't be very pronounced. But why `magic_*` should be isolated? I don't expect it to do `popen()` or something similar internally.

PR #220, opened 2016-10-21
use strncasecmp for non WIN32 platforms

alegen:
this allows case insensitive autocomplete on unix platforms


alegen:
Link to submitted issue which gives a better explanation - https://github.com/vifm/vifm/issues/221

xaizek:
`strnoscmp()` is a generic function, which is used for more than just completion. It's meant to do case sensitive comparison on *nix and be case insensitive on Windows, the way it's now.

Moreover this change will break file renames and some other functionality.

PR #270, opened 2017-05-04
Ability to format a name of a file moved to the trash directory

alegen:
Hello @xaizek,

For some time I have been meaning to add the ability of formatting the name of files moved to the trash directory. I needed this feature in order to better integrate my vifm workflow with some other scripts I made for myself along the years. Maybe some other people will find this useful. I would like to integrate these features upstream.

I am opening this pull request and am asking for your feedback. I tried to follow the same format that you are using, but am sure it is not 100% correct. Let me know what/how you would like to see improved in order to integrate the code into the official repo.

Also, I have ran some tests on linux and they seem to work well.

### DIFF
--- a/src/opt_handlers.c
+++ b/src/opt_handlers.c
@@ -195,6 +195,8 @@ static void timeoutlen_handler(OPT_OP op, optval_t val);
 static void title_handler(OPT_OP op, optval_t val);
 static void trash_handler(OPT_OP op, optval_t val);
 static void trashdir_handler(OPT_OP op, optval_t val);
+static void trash_filefmt_zero_handler(OPT_OP op, optval_t val);
+static void trash_filefmt_more_handler(OPT_OP op, optval_t val);
 static void tuioptions_handler(OPT_OP op, optval_t val);
 static void undolevels_handler(OPT_OP op, optval_t val);
 static void vicmd_handler(OPT_OP op, optval_t val);
@@ -675,6 +677,14 @@ options[] = {
 	{ "trashdir", "", "path to trash directory",
 	  OPT_STRLIST, 0, NULL, &trashdir_handler, NULL,
 	  { .init = &init_trashdir },
+	},
+	{ "trash_filefmt_zero", "", "file name when moved to the trash directory and no other files exist with the same name",
+	  OPT_STR, 0, NULL, &trash_filefmt_zero_handler, NULL,
+	  { .ref.str_val = &cfg.trash_filefmt_zero },
+	},
+	{ "trash_filefmt_more", "", "file name when moved to the trash directory and another file exists with the same name",
+	  OPT_STR, 0, NULL, &trash_filefmt_more_handler, NULL,
+	  { .ref.str_val = &cfg.trash_filefmt_more },
 	},
 	{ "tuioptions", "to", "TUI look tweaks",
 	  OPT_CHARSET, ARRAY_LEN(tuioptions_vals), tuioptions_vals,
@@ -2809,6 +2819,24 @@ trashdir_handler(OPT_OP op, optval_t val)
 		set_option("trashdir", val, OPT_GLOBAL);
 	}
 	free(expanded_path);
+}
+
+static void
+trash_filefmt_zero_handler(OPT_OP op, optval_t val)
+{
+	if (cfg.trash_filefmt_zero != NULL)
+		free(cfg.trash_filefmt_zero);
+	cfg.trash_filefmt_zero = malloc(strlen(val.str_val) + 1);
+	strcpy(cfg.trash_filefmt_zero, val.str_val);
+}
+
+static void
+trash_filefmt_more_handler(OPT_OP op, optval_t val)
+{
+	if (cfg.trash_filefmt_more != NULL)
+		free(cfg.trash_filefmt_more);
+	cfg.trash_filefmt_more = malloc(strlen(val.str_val) + 1);
+	strcpy(cfg.trash_filefmt_more, val.str_val);
 }
 
 /* Parses set of TUI flags and changes appearance configuration accordingly. */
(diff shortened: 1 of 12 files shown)


xaizek:
Hello,

Thanks for the contribution.

I see some inconsistencies (style, 'option'/%macro names, 'option' kind), change of default behavior, memory leak and breakage of file name "fixing" when moving files manually out of the trash. Also I think that amount of changes can be reduced significantly by reusing existing code (at the cost of using single-character macros like `%c`). 

I can point out those things in the pull request or just rework it if you don't mind me butchering it. Trash file names is probably the last thing that I expected to be configured, but apart from couple scenarios I don't see where it can go wrong and don't mind adding support for it.

alegen:
Hey @xaizek,

Thank you for the feedback. I would like to take another try an improve the implementation. Not that I mind you butchering it (hehe), but I want to provide something of better quality. Also I guess you have enough things to work on.

I would like to ask for more information about some of the points you mentioned:

1. change of default behavior - you mean here that the default format is not the same as before the patch?
2. breakage of file name "fixing" when moving files manually out of the trash - not sure i follow here, i tested the "undo" feature with deleted files and it seemed to work ok, but I guess this is not what you meant?
3. for the existing functionality, of the top of my head I will follow the implementation for the statusline option - is that a good approach?

Regarding the memory leak I will keep an open eye.

src/opt_handlers.c
```diff
@@ -2811,6 +2821,24 @@ trashdir_handler(OPT_OP op, optval_t val)
 	free(expanded_path);
 }
 
+static void
+trash_filefmt_zero_handler(OPT_OP op, optval_t val)
+{
+	if (cfg.trash_filefmt_zero != NULL)
+		free(cfg.trash_filefmt_zero);
+	cfg.trash_filefmt_zero = malloc(strlen(val.str_val) + 1);
+	strcpy(cfg.trash_filefmt_zero, val.str_val);
```
xaizek: These four lines are equivalent to:
```c
replace_string(&cfg.trash_filefmt_zero, val.str_val);
```

src/opt_handlers.c
```diff
@@ -2811,6 +2821,24 @@ trashdir_handler(OPT_OP op, optval_t val)
...
+}
+
+static void
+trash_filefmt_more_handler(OPT_OP op, optval_t val)
+{
+	if (cfg.trash_filefmt_more != NULL)
+		free(cfg.trash_filefmt_more);
+	cfg.trash_filefmt_more = malloc(strlen(val.str_val) + 1);
+	strcpy(cfg.trash_filefmt_more, val.str_val);
```
xaizek: Same here:
```c
replace_string(&cfg.trash_filefmt_more, val.str_val);
```

src/opt_handlers.c
```diff
@@ -676,6 +678,14 @@ options[] = {
 	  OPT_STRLIST, 0, NULL, &trashdir_handler, NULL,
 	  { .init = &init_trashdir },
 	},
+	{ "trash_filefmt_zero", "", "file name when moved to the trash directory and no other files exist with the same name",
```
xaizek: Neither Vim nor Vifm has options with underscores, I'd prefer to keep it that way for consistency.

src/opt_handlers.c
```diff
@@ -676,6 +678,14 @@ options[] = {
...
+	  OPT_STR, 0, NULL, &trash_filefmt_zero_handler, NULL,
+	  { .ref.str_val = &cfg.trash_filefmt_zero },
+	},
+	{ "trash_filefmt_more", "", "file name when moved to the trash directory and another file exists with the same name",
```
xaizek: I think Vim has too many options, which makes it sometimes hard to find those you need. I'd have a single option with syntax akin to:
```vim
set trashfilefmt=zero:something,more:somethingelse
```
Like in `'fillchars'` option (`fillchars_handler()`).

xaizek:
> Not that I mind you butchering it (hehe), but I want to provide something of better quality.

OK, I just didn't like to be asked to rewrite the bigger part of my patches and hate to do the same to other people. It might be a good idea to ping me before doing significant changes/additions, so that I could point out what's already present in the code base.

> change of default behavior - you mean here that the default format is not the same as before the patch?

Yes.  I'm not sure if anyone is relying on current format though, so it might be OK.

> breakage of file name "fixing" when moving files manually out of the trash - not sure i follow here, i tested the "undo" feature with deleted files and it seemed to work ok, but I guess this is not what you meant?

There is [`get_real_name_from_trash_name()`](https://github.com/vifm/vifm/blob/f3ccc979910c905891eeabdec48863e9b1a308ba/src/trash.c#L861-L878) function which relies on format. It needs to look up `trash_entry_t` similarly to the way it queries `get_resident_type(trash_path)` and use `trash_entry_t::path` to do not depend on trash-name format.

> for the existing functionality, of the top of my head I will follow the implementation for the statusline option - is that a good approach?

There is [`expand_custom_macros()`](https://github.com/vifm/vifm/blob/f3ccc979910c905891eeabdec48863e9b1a308ba/src/macros.h#L79-L80) which can be used like [here](https://github.com/vifm/vifm/blob/f3ccc979910c905891eeabdec48863e9b1a308ba/src/menus/apropos_menu.c#L44-L64).

> Regarding the memory leak I will keep an open eye.

You leaked `strmap_t *map` in `gen_trash_name`.

P.S. Didn't have a chance to use GitHub's review, tried it here just to see how it works.

src/opt_handlers.c
```diff
@@ -2811,6 +2821,24 @@ trashdir_handler(OPT_OP op, optval_t val)
 	free(expanded_path);
 }
 
+static void
+trash_filefmt_zero_handler(OPT_OP op, optval_t val)
+{
+	if (cfg.trash_filefmt_zero != NULL)
+		free(cfg.trash_filefmt_zero);
```
xaizek: By the way, all C standards explicitly allow argument of `free()` to be `NULL`, so such checks are never necessary.

alegen:
Hey @xaizek,

Sorry for taking longer to answer, was bussy travelling the past few days.
Thank you for the reply and the hints. I will start with them and submit another pull request when I feel ready. I will close this particular pull request.

PR #282, opened 2017-08-19
Fix directory size in statusline

dimonomid:
Statusline could only show 4K as a directory size, independently of
whether we already calculated the real dir size.


xaizek:
Thanks!

PR #285, opened 2017-08-19
Fix directory size for non-empty 'statusline'

dimonomid:


dimonomid:
My previous commit was not complete indeed. :)

Actually, `get_fentry_size` is not specific to statusline, so it should probably be factored out into more generic place and used across the rest of the codebase, but having little context of the project so far, I keep this change localized to statusline.

xaizek:
Looks like I failed to test it properly since I use `%E` macro. By the way, it's right below case statement for `%s` and uses already existing `get_file_size_by_entry()` function... I merged it way too fast :)

dimonomid:
Good point; I missed `get_file_size_by_entry()`. Updated.

xaizek:
Thanks!

PR #301, opened 2018-01-03
Correctly change back current directory after view update

d-e-s-o:
The file browser has the nice ability of changing the current working
directory to the currently displayed one. That changing happens
depending on which view is active, i.e., the left or the right side.
Unfortunately, if there are updates happening on the other side, for
example because a file increased its size, the current working directory
will be changed and not reverted as part of the view update process.
That is a problem when spawning off a new shell and expecting it to have
its working directory set to the currently active view's one.
This change fixes the problem by correctly restoring the current working
directory after each update operation.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -113,3 +113,6 @@ Dmitry Frank (a.k.a. dimonomid) fixed displaying size in statusline and
 implemented :histnext/:histprev commands.
 
 John Shea (a.k.a. coachshea) provided fix for the plugin related to neovim.
+
+Daniel Mueller fixed CWD of the process not matching current view after vifm
+picked up change in file system.
--- a/ChangeLog
+++ b/ChangeLog
@@ -169,6 +169,9 @@
 
 	Fixed the plugin in neovim.  Patch by John Shea (a.k.a. coachshea).
 
+	Fixed CWD of the process not matching current view after vifm picked up
+	change in file system.  Patch by Daniel Mueller.
+
 0.9-beta to 0.9
 
 	Escape $ and ` in %" macros on *nix systems.  Thanks to filterfalse.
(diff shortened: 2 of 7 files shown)


xaizek:
Thanks for the pull request. I understand how the issue could happen, but don't seem to able to reproduce it. Here's how I do it:
1. Open panes with two different directories
2. Put cursor in the right one
3. Run `for i in {1..100}; do usleep 100000 && rm bbbbbb && touch bbbbbb; done` in background at the location of the left pane.
4. Try to start shell from vifm.

And current directory is always the same. Could you explain how you manage to reproduce it?

d-e-s-o:
@xaizek I am at 192ef5c8dfb3173ff01f4d687315d5ac252cd875. In my left pane I am at `/tmp/` on my right pane somewhere else, call it `foo`. I am actually running vifm inside tmux. When I split the tmux pane the newly spawned shell's working directory is `foo`.

Now in another shell I change to `/tmp/` and run
```bash
while true; do echo hallo >> blubb; sleep 1; done
```

Switching back to tmux and splitting the tmux pane with vifm's right view active the shell's working directory is now `/tmp/`, and not `foo`.

With my change the newly spawned shell's working directory is `foo`.

d-e-s-o:
Are these Travis test failures expected? They seem to occur without my change as well.

xaizek:
@d-e-s-o, thanks for the explanation, I can see it now. tmux reads cwd directly from `/proc/` and this is how it's able to see what otherwise isn't available for external applications.

> Are these Travis test failures expected? They seem to occur without my change as well.

No, CIs are normally green. Recursive rename implicitly relied on cwd instead of using absolute paths, hence the failures. I already have a patch locally and will finish it up tomorrow, thanks.

d-e-s-o:
@xaizek Thank you for the fix!

PR #306, opened 2018-02-02
Always place cursor on top for newly opened dirs

dimonomid:
Otherwise, consider this:

- Open directory `/foo/bar`
- Move cursor to some non-first item
- Type `:cd ..` (so that we'll end up in `/foo`)
- Hit Enter (so that we'll end up in `/foo/bar` again)

At this point, cursor should be on the top. Without the proposed change,
it isn't.

I guess it's a very minor thing for the majority, but I constantly
change directories via external means, so my directory history is
"polluted" by random cursor positions, and it's really annoying when I
open a dir and expect the cursor to be on top, while it isn't.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,10 @@
 0.9.1-beta to current
 
+	Fixed opening a directory for which the saved cursor position is not at
+	the top: regardless of the cursor position we might have in the history,
+	when user hits Enter on a directory, the cursor should be placed on top.
+	Patch by Dmitry Frank (a.k.a. dimonomid).
+
 	Fixed :clone and :copy refusing to copy broken symbolic links, even though C
 	and p keys copy them.
 
--- a/src/running.c
+++ b/src/running.c
@@ -741,6 +741,11 @@ open_dir(view_t *view)
 	if(cd_is_possible(full_path))
 	{
 		navigate_to(view, full_path);
+
+		/* When opening a directory (effectively by hitting Enter on it), the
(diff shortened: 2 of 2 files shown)


dimonomid:
I'm not actually sure this is the right way to solve it, but it does solve it for me. Let me know if it should be done differently.

Also, I'm not sure I'm supposed to edit `AUTHORS` to add every single bit I've done in vifm, so I didn't. :)

xaizek:
This is a feature, not a bug :)

I guess a more fine-grained control over when cursor position is restored is needed, because while you want to do this less often, other people want to do this even more often that it happens now.

Actually, there was an attempt to do something similar with `'autochpos'` option, but it's too specific and can't address all possible scenarious. Instead of a boolean option, we probably need a list of cases when cursor should be restored (e.g., `cd`, `mark`, `bmark`, `open`).

dimonomid:
Thanks for clarification. Yeah, a list of cases sounds good.

I suspected that it's a feature actually, but then thought - no way. Well.. :) Tastes are different. I guess I'll keep that dirty hack for myself until a proper solution is implemented.

xaizek:
Added `'histcursor'` option. `direnter` value is responsible for the behaviour you want to suppress, but since some new cases when cursor is automatically positioned were added, you probably want to do
```vim
set histcursor-=startup,dirmark,direnter
```

dimonomid:
Thank you, that's awesome! :)

Issue #304, opened 2018-02-02
Show current sorting mode (or ideally just an arbitrary expression) in the status line


dimonomid:
I realized I miss the current sorting mode in the statusline, like `+name` or `+mtime`, etc. We could probably add one more macro to the list of supported macros to `statusline`, but I think it would be better to support arbitrary vifm expressions, to get much more flexibility, like `%{<expression>}`. Then I'd use `%{&sort}` and that's it.

What do you think?

xaizek:
Sounds like a reasonable request and I agree that `%{...}` is a nice way of doing this. There is no point in adding a new macro for things which can be evaluated via expressions.

dimonomid:
Ok, cool!

PR #307, opened 2018-02-04
Add support for arbitrary exps in statusline

dimonomid:
The syntax is: `%{...}`. E.g. `%{&sort}` evaluates to the value of the
`sort` option.

Resolves https://github.com/vifm/vifm/issues/304

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -4174,6 +4174,8 @@ since it cannot be selected
 .IP \- 2
 %z - short tips/tricks/hints that chosen randomly after one minute period
 .IP \- 2
+%{<expr>} - evaluate arbitrary vifm expression '<expr>', e.g. '&sort'
+.IP \- 2
 all 'rulerformat' macros
 .RE
 
@@ -4183,7 +4185,7 @@ minimum field width if you want field to be right aligned.
 Example:
 
 .EX
-  set statusline="  %t%= %A %10u:%\-7g %15s %20d "
+  set statusline="  %t%= %A %10u:%\-7g %15s %20d %{&sort} "
 .EE
 
 On Windows file properties include the following flags (upper case means flag is
--- a/src/ui/statusline.c
+++ b/src/ui/statusline.c
@@ -33,6 +33,8 @@
 #include "../compat/pthread.h"
 #include "../compat/reallocarray.h"
 #include "../engine/mode.h"
+#include "../engine/parsing.h"
+#include "../engine/var.h"
 #include "../modes/modes.h"
 #include "../utils/fs.h"
 #include "../utils/log.h"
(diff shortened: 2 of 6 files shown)


data/man/vifm.1
```diff
@@ -4174,6 +4174,8 @@ since it cannot be selected
 .IP \- 2
 %z - short tips/tricks/hints that chosen randomly after one minute period
 .IP \- 2
+%{<expr>} - evaluate arbitrary vifm expression '<expr>', e.g. '&sort'
```
xaizek: `vifm/data/vim/doc/app/vifm-app.txt` also needs to be updated accordingly.

src/ui/statusline.c
```diff
@@ -373,6 +375,48 @@ parse_view_macros(view_t *view, const char **format, const char macros[],
...
+					const char *e = strchr(*format, '}');
+					char *expr = NULL, *resstr = NULL;
+					var_t res = var_false();
+					ParsingErrors parsing_error;
+
+					/* If there's no matching closing bracket, add the opening one
+					 * literally */
+					if (e == NULL) {
+						strappendch(&result, &len, '%');
+						strappendch(&result, &len, '{');
```
xaizek: Wouldn't setting `ok = 0` do the same?

src/ui/statusline.c
```diff
@@ -373,6 +375,48 @@ parse_view_macros(view_t *view, const char **format, const char macros[],
...
+					parsing_error = parse(expr, &res);
+					if(parsing_error == PE_NO_ERROR) {
+						resstr = var_to_string(res);
+					}
+
+					if (resstr != NULL) {
+						copy_str(buf, sizeof(buf), resstr);
+					} else {
+						copy_str(buf, sizeof(buf), "<Invalid expr>");
+					}
```
xaizek: Style nitpicks to keep the style consistent (in tests too):
* should be no space after `if`
* curly braces need to be on separate lines

data/man/vifm.1
dimonomid: Done.

src/ui/statusline.c
dimonomid: Right, brainfart.

src/ui/statusline.c
dimonomid: Done.
However maintaining various style guidelines manually is annoying (especially when one got used to different conventions), I think it would be cool to employ clang-format, with the config which works for you, and pre-commit hook which would do all the magic. Combined with the CI job which would apply the clang-format and fail if there are differences, it would even eliminate the need to manually check style :)

dimonomid:
PTAL (I also updated AUTHORS as you suggested).

xaizek:
Thanks! I'll merge the changes after the release, don't want to introduce new changes in the last moment.

src/ui/statusline.c
xaizek: I'm not sure if clang-format is capable of handling the style, it's oriented towards big companies and their style guides. On the other hand this might be something to try to find out (maybe astyle will work), because I sometimes make style mistakes too :)

dimonomid:
Cool, thank you!

PR #310, opened 2018-02-05
Fix reported free space on *nix

dimonomid:
Make it consistent with what `df` reports.


dimonomid:
I noticed that vifm reports more free space than other tools (`df` and double commander), e.g. 18GB instead of 14GB. So here's the fix: use `f_bavail` (free blocks for unprivileged users) instead of `f_bfree` (free blocks).

xaizek:
Thanks!

PR #335, opened 2018-06-10
Remove the leading space from timefmt

cdlscpmv:
I still fail to understand why it was put there in the first place.

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -4380,7 +4380,7 @@ Number of spaces that a Tab in the file counts for.
 .BI 'timefmt'
 type: string
 .br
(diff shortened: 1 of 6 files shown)


xaizek:
Thanks.

> I still fail to understand why it was put there in the first place.

It could have been added to the value to make sure that all uses have the leading space (the second one is in status line formatting), but I don't actually remember. It definitely shouldn't have leaked to documentation.

Adding it on formatting probably was to prevent clashing with long file names. At that time there were no `'viewcolumns'` option, so controlling formatting in any way just wasn't available. Maybe it's worth dropping it completely now.

PR #358, opened 2018-07-29
Extend vifmrc with more examples (filetypes, fileviewer and commands)

jose1711:


xaizek:
Hm, not sure what to do with this. It's quite an overhaul and it breaks things. Those files are supposed to be installed on `make install` and vifm copies only a single file into `~/.vifm` automatically, so it won't work as is. However it could work if there were an environment variable like `$VIFMDATA` or something like that that expands to `/usr/share/vifm`.

jose1711:
Valid point, I did not think about integration with `make` TBH. I guess you can close the PR and I'll make a new one when "the time is right".

xaizek:
If it will be separated into more than one file, then something like `$VIFMDATA` is necessary to make it work without making vifm copy those files too, so it might be a prerequisite. Adding it to build system is easy, it's just editing of one `.am`-file.

LamprosPitsillos:
Why is this PR not closed?

It has a couple of useful things in it for new users.
Maybe open the Discussion tab on git hub and move some of it there?
Also having a discussion tab , in my opinion , greatly improves the user experience and helps move questions from the issues to the discussions.
What do you think?

xaizek:
There is [q2a](https://q2a.vifm.info/) site for that and [wiki](https://wiki.vifm.info/index.php/Main_Page) for useful things. Those who want to stay on GitHub can use issues, discussions tab doesn't seem to add anything useful except for a different UI.

PR #384, opened 2018-11-25
Set correct working directory on startup

d-e-s-o:
When vifm is started and it is configured to restore the previous
directory state (via set vifminfo=...,dhistory,savedirs,...), it will
set the program's current working directory according to that state.
However, the working directory ultimately set last is that of the right
file view, while the actually active file view is that on the left. That
causes confusion when forking off another program that will now have the
wrong working directory set.
To fix this problem this change reorders the auto-command triggering
code to first happen for the right view and only then for the left one.

### DIFF
--- a/src/event_loop.c
+++ b/src/event_loop.c
@@ -103,6 +103,10 @@ event_loop(const int *quit)
 	curr_input_buf = &input_buf[0];
 	curr_input_buf_pos = &input_buf_pos;
 
+	/* Make sure to set the working directory once in order to have the
+	 * desired state even before any events are processed. */
+	(void)vifm_chdir(flist_get_dir(curr_view));
+
 	while(!*quit)
 	{
 		wint_t c;


xaizek:
> [...] while the actually active file view is that on the left.

By default, but it can be the right pane if `'vifminfo'` contains `tui`. So relying on `curr_view` would be better if that code needs to be changed.

I'm not sure the change is necessary because of this line in `event_loop()` which is invoked shortly after the code you've reordered:

https://github.com/vifm/vifm/blob/de272ae4f1025370bdfbcfc7155e280231d9a7db/src/event_loop.c#L168

So I'd expect current directory to be correct on startup, however it might depend on the context in which a process is forked. What command is used after startup?

xaizek:
> Perhaps that's the event loop's `chdir`?

Yes, and it is reached after first key is pressed.

> Or do you have a better suggestion?

Since we need to cover state in which `vifm` awaits for input, I'd put `vifm_chdir()` call right above https://github.com/vifm/vifm/blob/de272ae4f1025370bdfbcfc7155e280231d9a7db/src/event_loop.c#L105-L107

We probably don't need to do this twice on every iteration, so calling it before entering the loop should suffice. This is more generic than in `vifm_main()` because `event_loop()` is also called from dialogs, which could potentially have similar issue in some cases.

d-e-s-o:
> > Or do you have a better suggestion?
> 
> Since we need to cover state in which `vifm` awaits for input, I'd put `vifm_chdir()` call right above
> 
> [vifm/src/event_loop.c](https://github.com/vifm/vifm/blob/de272ae4f1025370bdfbcfc7155e280231d9a7db/src/event_loop.c#L105-L107)
> 
> Lines 105 to 107 in [de272ae](/vifm/vifm/commit/de272ae4f1025370bdfbcfc7155e280231d9a7db)
> 
>   
>  while(!*quit) 
>  { 
> We probably don't need to do this twice on every iteration, so calling it before entering the loop should suffice. This is more generic than in `vifm_main()` because `event_loop()` is also called from message dialogs, which could potentially have similar issue in some cases.

Done!

xaizek:
Not like that :) Maybe I worded it poorly. Now it isn't done on every iteration. I meant adding an extra `vifm_chdir` before the loop to cover the beginning of the first iteration.

d-e-s-o:
Ah, I did misunderstand your comment. Please take another look.

xaizek:
Looks good, thank you.

PR #435, opened 2019-05-23
Simplify image preview

mb720:
Hi and thanks a lot for vifm!

The change simplifies the command to preview images and makes the output less redundant.

Instead of

    example.png PNG 1366x768 1366x768+0+0 8-bit sRGB 176847B 0.030u 0:00.026
    example.png=>/dev/null PNG 1366x768 1366x7 68+0+0 8-bit sRGB 176847B 0.200u 0:00.130

the output's now

    examlpe.png PNG 1366x768 1366x768+0+0 8-bit sRGB 176847B 0.000u 0:00.000


xaizek:
Hello. Thanks!

PR #441, opened 2019-06-22
Improve media menu

zsugabubus:
- Spawn media program in foreground,
- Show decorators for devices and mount-points,
- Make actions to be more human-friendly in regards of selected item,
- Accept `info=` for showing arbitrary text next to devices,
- Add [, ] keybindings for navigating between devices,
- Show blank lines only if there is enough space for them.

I didn't want to pollute the code with new commands like "eject"/"detach"/"power off" with random keybindings, instead user can extend her media program and implement these functionalities in that.
So running it in foreground solves this and my password input issues too.

I kept blank lines only if there is enough rows, so they won't be annoying if other devices go off-screen because of them; otherwise they help visually separating devices (as before). Plus, I added decorators also for visual appearance.

### DIFF
--- a/src/menus/media_menu.c
+++ b/src/menus/media_menu.c
@@ -206,7 +206,7 @@ reload_list(menu_data_t *m)
 	free(cmd);
 
 	int i;
-	const int menu_rows = (getmaxy(menu_win) + 1) - 3;
+	const int menu_rows = menu_win != NULL ? (getmaxy(menu_win) + 1) - 3 : 100;
 	int max_rows_with_blanks = info_count       /* Device lines. */
 		                       + (info_count - 1) /* Blank lines. */;
 	for(i = 0; i < info_count; ++i)
--- a/src/menus/media_menu.c
+++ b/src/menus/media_menu.c
@@ -27,41 +27,47 @@
 #include "../compat/fs_limits.h"
 #include "../compat/reallocarray.h"
 #include "../modes/dialogs/msg_dialog.h"
-#include "../ui/cancellation.h"
+#include "../ui/ui.h"
 #include "../ui/statusbar.h"
 #include "../utils/fs.h"
 #include "../utils/path.h"
 #include "../utils/str.h"
 #include "../utils/string_array.h"
 #include "../utils/utils.h"
-#include "../background.h"
+#include "../running.h"
 #include "menus.h"
 
 /* Example state of the menu:
  *
  *    Items                              Data            Action
  *
- *    /dev/sdc
- *     -- not mounted                    m/dev/sda3      mount /dev/sda3
+ *    /dev/sdc                                           mount /dev/sda3
+ *    `-- (not mounted)                  m/dev/sda3      mount /dev/sda3
  *
- *    /dev/sdd [label]
- *     -- /                              u/              unmount /
- *     -- /root/tmp                      u/root/tmp      unmount /root/tmp
+ *    /dev/sdd {text}
+ *    |-- /                              u/              unmount /
+ *    `-- /root/tmp                      u/root/tmp      unmount /root/tmp
+ *
+ *    /dev/sde                                           unmount /media
+ *    `-- /media                         u/media         unmount /media
  *
  * Additional keys:
  *  - r -- reload list
+ *  - [ -- previous device
+ *  - ] -- next device
  *  - m -- mount/unmount
  *
  * Behaviour on Enter:
- *  - for lines with mount-point: navigate inside it
- *  - for all other lines: just close the menu
+ *  - navigate inside mount-point or,
+ *  - mount device if there are no mount-points or,
+ *  - do nothing and keep menu open
  */
 
 /* Description of a single device. */
 typedef struct
 {
 	char *device;   /* Device name in the system. */
-	char *label;    /* Label of the device (can be NULL). */
+	char *text;     /* Arbitrary text next to device name (can be NULL). */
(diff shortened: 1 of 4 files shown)


zsugabubus:
Eh, I forgot to update tests. Does this segfault means that something is very bad with my code or this is how failed tests are indicated?

zsugabubus:
Test fails, because:
- I forgot to update them, and formatting differs in one-one chars,
- blanks lines hidden if screen is small and I guess "screen is small" when the tests were run (used `getmaxy` to determine).

I don't know what to do with latter. Maybe modify test and delete blank line tests?

xaizek:
Thank you! Although next time please implement separate changes in separate commits.

> Maybe modify test and delete blank line tests?

Something like this would be more deterministic:

```c
const int menu_rows = (menu_win == NULL ? 0 : (getmaxy(menu_win) + 1) - 3);
```

I might make update it to allow setting menu height from tests.

zsugabubus:
> Off-topic, but I wouldn't like to open an issue for it: running tests on unmodified code resulted in: `utils/utils_nix.c:681:26: runtime error: null pointer passed as argument 1, which is declared to never be null`. Adding `update_string(&cfg.slow_fs_list, "");` to `tests/misc/integration.c` fixed it.

I've modified the `execute_media_cb` to _do not_ close media menu if enter did nothing (e.g. we are on a blank line), so I returned 1 in that cases. However I get the stack trace below during testing. Changing back to `return 0` works well (at least it does not show stack trace and works in release mode).

```c
TEST(enter_does_nothing_on_device_lines_with_multiple_mounts)
{
	FILE *fp = fopen("script", "w");
	fputs(SHEBANG_ECHO "\n\
device=/dev/sdf\n\
label=sdf-label\n\
mount-point=sdf-mount-point\n\
mount-point=jkl-mount-point\n", fp);
	fclose(fp);

	assert_success(exec_commands("media", &lwin, CIT_COMMAND));

	lwin.curr_dir[0] = '\0';
/*185*/ (void)vle_keys_exec(WK_CR);
	assert_string_equal("", lwin.curr_dir);
}
```
(Slightly modified the script writing to decrease noise.)

```
==12986==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000074 (pc 0x7f4a7f137a4e bp 0x55ff74c5ef20 sp 0x7fff2281a768 T0)
==12986==The signal is caused by a READ memory access.
==12986==Hint: address points to the zero page.
    #0 0x7f4a7f137a4d in is_term_resized_sp (/usr/lib/libncursesw.so.6+0x3aa4d)
    #1 0x55ff73ea6044 in update_term_size ui/ui.c:1297
    #2 0x55ff73ea6b78 in ui_update_term_state ui/ui.c:361
    #3 0x55ff73ea7031 in resize_for_menu_like ui/ui.c:1237
    #4 0x55ff73e915b8 in menus_full_redraw menus/menus.c:319
    #5 0x55ff73e7532b in execute_mapping_handler engine/keys.c:1198
    #6 0x55ff73e7593b in dispatch_key engine/keys.c:671
    #7 0x55ff73e75e77 in execute_next_keys engine/keys.c:637
    #8 0x55ff73e76442 in dispatch_keys_at_root engine/keys.c:444
    #9 0x55ff73e765c6 in dispatch_keys engine/keys.c:341
    #10 0x55ff73e766e1 in execute_keys_general engine/keys.c:308
    #11 0x55ff73e769f6 in execute_keys_general_wrapper engine/keys.c:288
    #12 0x55ff73e769f6 in vle_keys_exec engine/keys.c:257
    #13 0x55ff73df9544 in enter_does_nothing_on_device_lines_with_multiple_mounts misc/menus_media.c:185
    #14 0x55ff73df8208 in stic_fixture stic/stic.h:351
    #15 0x55ff73e1dc0d in stic_suite misc/suite.c:18
    #16 0x55ff73e3542d in run_tests stic/stic.c:531
    #17 0x55ff73e36aa3 in stic_testrunner stic/stic.c:670
    #18 0x55ff73d549fa in main misc/suite.c:18
    #19 0x7f4a7e5a8ee2 in __libc_start_main (/usr/lib/libc.so.6+0x26ee2)
    #20 0x55ff73d54f9d in _start (/run/user/1000/memory/vifm/tests/bin/misc+0xc14f9d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/lib/libncursesw.so.6+0x3aa4d) in is_term_resized_sp
==12986==ABORTING
```

xaizek:
> Off-topic, but I wouldn't like to open an issue for it:

Thanks.  Fill free to commit the change (as a separate commit).  I don't get that error probably due to compiler differences.

> However I get the stack trace below during testing.

It now reaches code that was never executed in tests before.  It doesn't crash for me, but the usual thing to do (if there is no better way like here) is to disable behaviour for tests.  Try applying this:

```patch
diff --git a/src/ui/ui.c b/src/ui/ui.c
index f81be1ee1..658e464cd 100644
--- a/src/ui/ui.c
+++ b/src/ui/ui.c
@@ -1278,6 +1278,12 @@ ui_setup_for_menu_like(void)
 static void
 update_term_size(void)
 {
+	if(curr_stats.load_stage < 1)
+	{
+		/* No terminal in tests. */
+		return;
+	}
+
 #ifndef _WIN32
 	struct winsize ws = { .ws_col = -1, .ws_row = -1 };
 
```

xaizek:
Thank you!

PR #442, opened 2019-06-23
Fix trash directory annoyances

zsugabubus:
* Create trash directory only if user wants it.
* If file system is read-only, do not warn about non-creatable trash.


xaizek:
Thanks!

PR #444, opened 2019-06-24
Improve performance of tree prefix generation

zsugabubus:
* ~Slightly improves reversing algorithm (shouldn't be so significant, but makes at least 4 times faster),~
* Doesn't require hardcoded strings to be reversed.

### DIFF
--- a/src/ui/fileview.c
+++ b/src/ui/fileview.c
@@ -1346,7 +1346,7 @@ mix_in_file_name_hi(const view_t *view, dir_entry_t *entry, col_attr_t *col)
 TSTATIC void
 format_name(int id, const void *data, size_t buf_len, char buf[])
 {
-	size_t len, i;
+	size_t len;
 	dir_entry_t *child, *parent;
 
 	const column_data_t *cdt = data;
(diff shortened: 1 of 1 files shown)


src/ui/fileview.c
```diff
@@ -1379,27 +1379,30 @@ format_name(int id, const void *data, size_t buf_len, char buf[])
...
 		(void)sstrappend(buf, &len, buf_len + 1U, prefix);
 
 		child = parent;
 		parent -= parent->child_pos;
 	}
 
-	for(i = 0U; i < len/2U; ++i)
+	/* Reversing can be performed on chunks of four chars. */
+	uint32_t *left  = (uint32_t *)(buf),
+	         *right = (uint32_t *)(buf + len);
```
xaizek: This seems to be asking for a UB. C99, §6.3.2.3:

```
7 A pointer to an object or incomplete type may be converted to a pointer to a
  different object or incomplete type. If the resulting pointer is not correctly 
  aligned for the pointed-to type, the behavior is undefined. [...]
```
We can't guarantee that `buf` is properly aligned as it comes from the outside and branching based on whether alignment is correct seems to be overkill.

Reversing can be completely avoided by printing from right to left, but that would need figuring out depth of a node first, which might take more time than reversing (I haven't measured it though).

zsugabubus: It didn't catch my attention, thanks.

Hopefully, it can be fixed if you think it's worth it. Formatting functions only get called here, so adding alignment specifiers could solve it.

https://github.com/vifm/vifm/blob/4307041c509613d3fbca58d17b25469a7857ea08/src/ui/column_view.c#L280-L286

xaizek: I don't think that `column_view.c` should know about implementation details of formatters. The unit is meant to not have such dependencies.

I might have a better solution without reversing or using extra memory, but if we are talking about optimizations, we should at least try to measure their effect. Did you do any measurements yet?

zsugabubus: Sorry for not taking measures before. It only means 3-4 milliseconds "speedup" for 10 thousands of lines (depth ~6-7), so it just doesn't worth the effort. Next time I will be much more careful.

PR #454, opened 2019-07-19
Simplify postponed view updating

zsugabubus:
Time was only needed to provide a monotonic counter, but if I'm right, simple boolean flags can get the job done, too.


xaizek:
> if I'm right, simple boolean flags can get the job done, too.

Yes, even counter isn't needed. It probably grew out of mutex not being there initially. Might have been an attempt to get away without one (disregarding the data race).

Thanks!

PR #455, opened 2019-07-19
Reset decoration cache on change

zsugabubus:
Decoration index cache must be reset for every dir_entry in both views
including **left and right columns**. Although, a full reload follows
the change, it can fire redraw events (before the actual reload happens)
in some circumstances (e.g. file list change) that potentially use
invalid decoration indexes from the cache.

It was really hard to find out how I can reproduce this, because at first I was more than sure that it has to be a data race. Surely not the simplest but a stable way to reproduce:

#### Reproduction
0) Assure that you have some items in `'classify'`.
1) Open two panes L and R in split view. `set millerview` in L and focus R.
2) Execute `:write | edit $MYVIFMRC | restart<CR>` (defaults to `,c`),  add `set classify=""` at the end, and most importantly delete a file (but not all!) from the directory of pane L (open another vifm or use terminal). It should cause a file list change in the inactive pane.
3) Close your config editor.
4) Vifm should crash.
(It can be reproduced using single pane, too.)

@xaizek  Could you please add tests? Probably, you could even make test case more simpler.

_Edit: deleted word + clarified reproduction_

### DIFF
--- a/src/opt_handlers.c
+++ b/src/opt_handlers.c
@@ -1662,20 +1662,11 @@ classify_handler(OPT_OP op, optval_t val)
 
 	if(str_to_classify(val.str_val, type_decs) == 0)
 	{
-		int i;
-
 		assert(sizeof(cfg.type_decs) == sizeof(type_decs) && "Arrays diverged");
 		memcpy(&cfg.type_decs, &type_decs, sizeof(cfg.type_decs));
 
-		/* Reset cached indexes for name-dependent type_decs. */
-		for(i = 0; i < lwin.list_rows; ++i)
-		{
-			lwin.dir_entry[i].name_dec_num = -1;
-		}
-		for(i = 0; i < rwin.list_rows; ++i)
-		{
-			rwin.dir_entry[i].name_dec_num = -1;
-		}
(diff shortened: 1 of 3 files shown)


xaizek:
> Could you please add tests? Probably, you could even make test case more simpler.

Sure, once I reproduce it :) (Wasn't successful at it yet.)

zsugabubus:
Hmm.

> File list change in inactive but visible miller view pane with matching `'classify'` in the left or right column.

1) `cd /path/to/vifm/sourcecode/tests/ && vifm --no-config `
2) `:nnoremap K :split<CR>:set classify+='X ::tests/::/' sortnumbers millerview<CR>j<TAB>:touch newfile|restart<CR>`
3) Press `K`. 

I don't know vifm's source code in depth, but I guess there are numerous other code paths where it can fail because it makes a redraw before the reload after `'classify'` had changed.

xaizek:
Thank you. The second sequence crashed immediately.

xaizek:
Merged as e9355d5b46dd05499a2f6db2fdcd78bdf074a87d. Thanks!

PR #489, opened 2019-10-30
reallocarray: avoid interposing when provided by the system

mathstuf:
---
Not sure how you want the autoconf stuff regenerated; I have 2.16 locally which makes a lot more diff noise around.

Fedora bug: https://bugzilla.redhat.com/show_bug.cgi?id=1480309


xaizek:
Thanks, I didn't know it was added to glibc, but I'd do the patch differently.

Your changes keep vifm's declaration even when system's `reallocarray()` is being used. If libc's implementation is in use, the declaration should also come from it.

I have pushed a commit to your branch in case you'd like to check it as my glibc is older than 2.16. However, I don't really expect things to go wrong there.

mathstuf:
Sure, that looks cleaner. Not much of an autotools guru myself :) . Tested locally to still work. Feel free to squash it down if you'd like.

xaizek:
Thanks for testing.

Issue #498, opened 2019-12-07
shellcmdflag can not be of arbitrary size on windows


basbossink:
I was trying to configure vifm in such a way that I could use a version of powershell (core) as the shell. Since I installed it as a dotnet global tool on my system the commandline needed to run a single command would be something like:
```cmd
dotnet C:\Users\bas\.dotnet\tools\.store\powershell\6.2.3\powershell\6.2.3\tools\netcoreapp2.1\any\win\pwsh.dll -nop -noni -nol -c
```
I could not get this to work properly so I decided to investigate further and look into the code.
In `src/utils/utils_win.c:198` I found:
```c
	char buf[strlen(cfg.shell) + 5 + strlen(cmd)*4 + 1 + 1];
```
looking at line 210-211:
```c
		const char *sh_flag = (by == SHELL_BY_USER ? cfg.shell_cmd_flag : "/C");
		snprintf(buf, sizeof(buf), "%s %s '", cfg.shell, sh_flag);
```
I came to the conclusion that the size calculation for the allocated `buf` _assummes_ the custom set `cfg.shell_cmd_flag` has a length of `2` like the default `/C`.

I guess this could be fixed be moving line 210 before line 198 and modifying it to something like:
```c
	char buf[strlen(cfg.shell) + 1 + strlen(sh_flag) + 2 + strlen(cmd)*4 + 1 + 1];
``` 

I would be happy to supply a PR for this but I can't seem to find in the current `README` or `HACKING.md` file how to build vifm on windows or maybe cross-compile it on Windows Subsystem for Linux (WSL).

Does my investigation make any sense to you or am I missing something?

xaizek:
You're right, the size of the buffer is too small. Thanks for the report and finding out where the bug is.

Instructions for building vifm on Windows are [on the wiki](https://wiki.vifm.info/index.php/Obtaining_Vifm#Windows). Building using MXE inside WSL might be easier than using MinGW. However, given the size of the change, I could just give you binaries for test once you submit a PR, so you don't have to spend too much time setting up build environment.

xaizek:
Fixed by #499.

PR #499, opened 2019-12-10
Fix issue #498

basbossink:
Include the actual size of the command flag in the size of the
buffer. Also simplified the code between the different cases.

Unfortunately I have not been able to test this, it compiles but
does not link using MXE on WSL, a MSYS2 build also fails.

### DIFF
--- a/src/utils/utils_win.c
+++ b/src/utils/utils_win.c
@@ -195,35 +195,28 @@ win_exec_cmd(char cmd[], int *returned_exit_code)
 char *
 win_make_sh_cmd(const char cmd[], ShellRequester by)
 {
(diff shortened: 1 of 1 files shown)


xaizek:
When I went to test your changes, I realized that using non-cmd on Windows was broken by introducing `'shellcmdflag'`, made a quick fix for it only now. I have also reverted you removing code for escaping, because I think it's necessary.

Here's a 64-bit binary: [vifm.exe.zip](https://github.com/vifm/vifm/files/3964099/vifm.exe.zip)

xaizek:
Will you be able to test if the binary works for your use case or should I assume that it does?

xaizek:
Thanks, merged it. After rebasing, the commit is a664df209cf018782637318e8e6e9309360718b8.

PR #501, opened 2020-01-01
Add script to convert vim color schemes to vifm

Quiark:
Simple script (for Vim) that will convert its color schemes to vifm format


xaizek:
Thanks! The output colors are obviously not perfect, but should be a good start for anyone willing to port a Vim scheme. I'll probably move it inside Vim plugin and add a :command to convert a scheme to make it easier to use.

PR #503, opened 2020-01-03
Fix wrong cursor position after initialization

d-e-s-o:
Change 1a5a0571c37cdc81fcfeb7cba763844cd9591320 broke the cursor
position on a freshly opened instance (and also after switching panels).
Previously, the cursor was always correctly positioned on the selected
entry. With said change it is located one line below.
I am not familiar with the code base but reordering the call to
`fview_position_updated` solves the problem.

### DIFF
--- a/src/ui/fileview.c
+++ b/src/ui/fileview.c
@@ -822,11 +822,6 @@ redraw_current_view(void)
 void
 fview_cursor_redraw(view_t *view)
 {
-	/* fview_cursor_redraw() is also called in situations when file list has
-	 * changed, so just let fview_position_updated() deal with it.  With a cache
-	 * of last position, it should be fine. */
-	fview_position_updated(view);
-
 	/* Always redrawing the cell won't hurt and will account for the case when
 	 * selection state of item under the cursor has changed. */
 	if(view == other_view)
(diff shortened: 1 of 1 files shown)


xaizek:
> broke the cursor position on a freshly opened instance (and also after switching panels).

Thanks for bisecting, but I don't see it being broken in these cases, can't reproduce it with switching panes in `vifm --no-configs` either (quit with `:quit!` to do not affect `vifminfo`). Tried both on `master` and on `0.10.1`.

I wonder what might be causing this in a configuration. Can you post your `vifmrc`? I want to try to bisect it.

d-e-s-o:
Let me clarify that when I say cursor I am talking about the terminal's cursor. I am specifically not referring to the selected line or anything in `vifm`'s UI directly.

The regression is only really noticeable if, for some reason, you rely on said cursor. For me that's the case because of `tmux`' selection feature.

I ruled out that it's a config issue by running with `--no-configs`. Perhaps this video makes it more clear (notice where the black cursor is located when the tab title changes from "vifm" to "[tmux]"):

output

xaizek:
Thanks for the clarification. Now that you said it's a terminal/hardware cursor, I can see it being off. I'll need to look into whether there is a better way of fixing this, since the cursor is set explicitly and I want to know why it's off by one.

xaizek:
Thanks for the report. I don't want to reverse order of the code and applied a different fix.

d-e-s-o:
Thanks for the quick fix! Works for me.

PR #523, opened 2020-03-05
Fixed typo in man page in Visual Mode section

maxigaz:


xaizek:
Thanks, could you also update `data/vim/doc/app/vifm-app.txt`?

maxigaz:
> Thanks, could you also update `data/vim/doc/app/vifm-app.txt`?

Oh, I haven't noticed that file.

(Now that I've pushed another commit, that should be it, right? I'm asking because I'm relatively new to pull requests.)

xaizek:
Yep, that's it, thank you.

PR #563, opened 2020-07-02
Add IO rate calculation

alborzjafari:
Displays IO rate when copying file(s).

### DIFF
--- a/src/fops_common.c
+++ b/src/fops_common.c
@@ -24,6 +24,8 @@
 #include <fcntl.h>
 #include <sys/stat.h> /* stat umask() */
 #include <sys/types.h> /* mode_t */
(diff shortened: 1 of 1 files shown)


xaizek:
Thank you for the contribution.

Are there specific reasons to do not use `friendly_size_notation()` to format the rate or you just didn't notice the function?

alborzjafari:
> Thank you for the contribution.
> 
> Are there specific reasons to do not use `friendly_size_notation()` to format the rate or you just didn't notice the function?

You're welcome.

I just preferred to use a function with a lower CPU load. Besides, I did a bit of optimization in the conditional statements, because MB/s happens more frequently.

Issue #634, opened 2021-01-20
Unable to qmap c-up/down, c-home/end ?


mckellygit:
Hi,

Thank you for vifm, it is awesome.
I have added many qmaps to my liking to match my vim and less preferences, and its great!
I cannot however figure out how to map:
\<c-up\>, \<c-down\>, \<c-home\>, \<c-end\>
I try these qmaps and they do not work.
I have tried the correct esc-codes as well but still nothing.
I am on Ubuntu 18.04 with the latest master vifm built with ENABLE_EXTENDED_KEYS set.
Any ideas ?  Do you think this possible without source changes ?

thx for any help,
-m

xaizek:
Hello,

If you can't find a particular angle-bracket notation in [the documentation](https://vifm.info/vimdoc.shtml#vifm-mappings), then it's not supported. curses almost doesn't expose keys with modifiers and I'm reluctant to hard-code them.

However, mapping escape codes should work. For `<c-up>`:

    :qnoremap <esc>[1;5A j

mckellygit:
Hi,
The esc-code mapping also does not work here.  I will double check my config.
But I've seen this before with (n)curses, when extended key support is enabled then <c-up> and others actually return a keycode above the KEY_MAX value, instead of the esc-code sequence.
I had to loop though the keyname()s of kDN5, kUP5, etc. to get the keycode for each, and use that to know what was pressed.
thx,
-m

xaizek:
Without `keypad(window, TRUE)` arrows and other grey keys don't work, so it's on unless you specify `--disable-extended-keys`:

https://github.com/vifm/vifm/blob/master/src/ui/ui.c#L256-L258

I haven't observed the behaviour you describe. Maybe that's an extension of a particular curses implementation?

mckellygit:
Yes, arrow keys work fine unless I build with --disable-extended-keys.
Its just ncurses-6.2
Does the esc-code mapping for \<c-down\> (\<esc\>[1;5B) work for you when you have --enable-extended-keys ?
extended-keys adds KEY_DOWN for \<down\> etc. - that is all fine, but [at least in my case] it doesn't have a specific keycode for \<c-down\> - and here is the rub, it still returns a 5000+ value (> KEY_MAX).  This is the case for several keys.

All I am saying is I can try:
1). build with --disable-extended-keys and add many esc-code mappings for \<up\>, etc. to my vifmrc
or
[...]

PR #636, opened 2021-01-22
Disable extended keys above KEY_MAX so that esc-codes can be used in mappings

mckellygit:
Hi,

Thank you for awesome vifm.
Here is a PR for issue https://github.com/vifm/vifm/issues/634 - if you find it useful.

thx,
-m

### DIFF
--- a/src/ui/ui.c
+++ b/src/ui/ui.c
@@ -215,6 +215,21 @@ setup_ncurses_interface(void)
 {
 	int screen_x, screen_y;
 
+#ifdef ENABLE_EXTENDED_KEYS
+#if defined(NCURSES_EXT_FUNCS) && NCURSES_EXT_FUNCS >= 20081102
+	/* Disable 'unsupported' extended keys so that esc-codes will be
+	 * received instead of unnamed extended keycode values (> KEY_MAX).
+	 * These keys can then be mapped with esc-codes in vifmrc.
+	 * An example could be <c-down>:
+	 *     qnoremap <esc>[1;5B j
+	 * as without this change <c-down> can return a keycode value of 531
+	 * (terminfo name kDN5), which is larger than KEY_MAX and has no
+	 * pre-defined curses key name.
+	 * NOTE: this MUST be called before initscr() */
+	use_extended_names(false);
+#endif
+#endif /* ENABLE_EXTENDED_KEYS */
+
 	initscr();
 	noecho();
 	nonl();


mckellygit:
Hi,
I tested this, but it does not work.  The esc-code mappings for \<c-down\> and some others now no longer work.
I used:
```
   use_extended_names(false);
```
instead of my loop of keyok(i, false) at the same location in the code.

mckellygit:
I just added a word to the comment

xaizek:
Last request: please try calling `use_extended_names(false);` before `initscr()`. If it doesn't work, then I guess it doesn't do what I think it does and it's of no use.

I've also found [an answer](https://stackoverflow.com/a/29730981/1535516) by Thomas Dickey, which talks about `define_key()` function. Maybe it could be used in the future to avoid mapping `<esc>...` sequences. (This is more of a note to myself.)

mckellygit:
Hi,

Yes! - calling it earlier, before initscr() works!

And yes, the harder way - we could have defined our own keys with define_key() -
#define KEY_CDOWN 5002
define_key("\033[1;5B", KEY_CDOWN); 
Even better is to query for keynames such as kDN5 etc and then you know the capability and dont have to hard-code any esc codes.
But the use_extended_names(false) is best.
thx,
-m

mckellygit:
Do you want me to update the PR or just close it ?

mckellygit:
Probably should wrap the call in an ifdef for the ncurses version/usage, as use_extended_names() might not be in all curses

xaizek:
Great! Thanks a lot for testing.

Yes, it needs a `NCURSES_EXT_FUNCS` `#ifdef` around it. Update the PR, no need to close it.

`kDN5` isn't a standard name, so looking it up doesn't sound right. Hard-coding will depend on a terminal also. I was thinking about adding a command like `:newkey <c-down> <esc>[1;5B` after which `qnoremap <c-down> ...` would make sense.

mckellygit:
Updated PR and squashed.
I like your two-step approach so esc codes are user/config-defined, it would be a good alternative/addition one day.
thx,
-m

xaizek:
Merged, thank you!

mckellygit:
Thanks.
Perhaps it should be 
#if defined(NCURSES_EXT_FUNCS) && defined(NCURSES_XNAMES)
but I dont think it really matters much.
-m

mckellygit:
@xaizek hi,
Just some info, we tried this same thing in tig, removing the keyok(false) code and calling use_extended_names(false) instead.
It works, but it was discovered that mouse clicks no longer work on a Mac with Terminal.  Mouse seems ok in some other terms though.
I just wanted to mention this if it maybe is useful one day.

xaizek:
Thanks, I'll update the comment just in case (I could never make the mouse work adequately).

mckellygit:
From koutcher -
The problem occurs with xterm (and derived) terminal descriptions. It seems to be linked to changes introduced in ncurses 6.1 regarding kmous capability.

Issue #659, opened 2021-04-10
Add keybindings to permissions dialog


qsmodo:
I believe the permissions dialog could benefit from some shortcuts for common operations or/and a layout change. For example, a single keypress {r,w,x} to set/unset all read/write/execute bits (normally I'd just like to do `cpx` to make a script executable), or/and maybe a 2D layout such as
```
       Read  Write  Execute  Special
Owner  [ ]    [ ]     [ ]      [ ]
Group  [ ]    [ ]     [ ]      [ ]
Other  [ ]    [ ]     [ ]      [ ]
```
so as to make navigation more convenient with {h,j,k,l} instead of only a vertical navigation with {j,k}.
I have suggested "Special" for SetUID, SetGID and Sticky but there may be some better description for that.

xaizek:
Sounds like a great suggestion that makes the dialog much more usable. Thanks!

PR #663, opened 2021-04-14
Add keybindings to permissions dialog

qsmodo:
I implemented keybindings to set/unset all the bits of a same class in the permission dialog (as requested #659). Namely, `r` sets/unsets all the r bits, and so on. Also `e` checks the "Set Recursively" flag.

I added some comments in some other parts of the file. Since I did not notice comments in that file, feel free to remove them, but I think they are somehow helpful.

Let me know in case there is something I overlooked or that I can improve. I have neither access to nor familiarity with Windows so those are for Unix only.

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -6012,6 +6012,16 @@ One shortcut per sorting key (see the dialog).
 h, Space \- check/uncheck.
 .br
 q \- close dialog
+.br
+r \- (*nix only) (un)set all read bits
+.br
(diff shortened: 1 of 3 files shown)


xaizek:
Thanks for the comments. You've updated man-page, but not `data/vim/doc/app/vifm-app.txt` and `cmd_recurse` should be called `cmd_e`.

xaizek:
Thanks!

PR #668, opened 2021-04-25
Add %v for vertical multiplexer split

qsmodo:
Vifm has the `%s` macro, but this always splits horizontally. For a vertically split window, I propose `%v`.

I have tested the patch both in Screen and Tmux. 

Let me know if there is something I can improve. Cheers!


xaizek:
Thanks, haven't tried it yet, but it looks good.

You could also extend [this test](https://github.com/vifm/vifm/blob/master/tests/misc/expand_macros.c#L281) for new macro.

xaizek:
Thanks!

Issue #514, opened 2020-02-16
Move directory and file count in preview to the top


AtomToast:
At the moment the amount of files and directories (for example ` 2 directories, 42 files` ) in the preview window of a selected directory (for example through pressing w, e or using :view) are located at the bottom of the tree view.  
This makes them disappear off the screen when looking at large directories which especially then would be a useful information to have.

xaizek:
Makes sense, thanks for the suggestion.

xaizek:
On second thought, it might be better to make this optional. Current implementation allows viewing top of huge trees fast because it quits upon filling window, but printing files and directories stats at the top will require visiting whole directory tree anyway.

AtomToast:
That makes sense I guess, I am all for options and configurability

PR #676, opened 2021-05-19
Allow file count before tree in quickview

qsmodo:
This is meant to address issue #514 with a new boolean option, `countfirst`, which allows one to move the count of directory and non-directory files to the top of the quick view when listing a directory. 

Handling the cache part was a bit challenging for me, so, as usual, please let me know if there is something to improve. Cheers!

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -4306,6 +4306,7 @@ view mode).
   item               default  meaning
   graphicsdelay:num  0        delay before drawing graphics (microseconds)
   hardgraphicsclear  unset    redraw screen to get rid of graphics
+  toptreestats       unset    show file counts before the tree in quick view
 
 graphicsdelay is needed if terminal requires some timeout before it can
 draw graphics (otherwise it gets lost).
--- a/data/vim/doc/app/vifm-app.txt
+++ b/data/vim/doc/app/vifm-app.txt
@@ -3605,6 +3605,7 @@ view mode).
     item               default  meaning ~
     graphicsdelay:num  0        delay before drawing graphics (microseconds)
     hardgraphicsclear  unset    redraw screen to get rid of graphics
+    toptreestats       unset    show file counts before the tree in quick view
 
 graphicsdelay is needed if terminal requires some timeout before it can
 draw graphics (otherwise it gets lost).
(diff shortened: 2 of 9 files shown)


qsmodo:
Sorry, I forgot to check for memory leaks and there are some. I'll let you know once I have solved them.

qsmodo:
It seems the tests have failed because I pedantically changed "0 directories" to "0 directory". Would you like me to revert that change?

xaizek:
> It seems the tests have failed because I pedantically changed "0 directories" to "0 directory". Would you like me to revert that change?

Yes, please. Not because of tests, but because it was correct [for English](https://english.stackexchange.com/questions/38293/why-is-zero-followed-by-a-plural-noun). Test failures also indicate that you broke collecting stats when `countfirst` is off. It now counts only top-level entries.

> a new boolean option, `countfirst`

There is already `previewoptions`. Please just add new value there. I'd pick something more specific, like `toptreestats`.

Implementing this isn't that straightforward. You basically have to not stop the traversal upon reaching height limit, but not print anything, just collect statistics. Then this information needs to be printed at the top somehow. Might have to do it this way:

1. Print, say, a line with 50 spaces at the beginning: `fprintf(fp, "%.50s\n", "");`
2. [stats at the top] At the end seek to the beginning, print stats there and seek to the start again.
3. [stats at the bottom] At the end seek to the position after that line (offset 51 in this case).

Also, would be nice to have a test for new functionality (in `tests/misc/view_dir.c`).

Thanks and let me know if you need help.

qsmodo:
Oh, I overlooked a couple of things then (no wonder how I did not notice it counted nested files too). I'll set out to fix those points, many thanks for your hints and time reviewing this.

qsmodo:
I have followed your suggestions and hopefully fixed the problems.

This time I have the left the directory traversing function, `print_dir_tree`, alone. Instead, my approach is to set `max_lines = INT_MAX` if `toptreestats` is set. By giving the window that height, it does not fill up and thus the listing doesn't exit prematurely.

xaizek:
Thanks! I've changed how stats are collected to improve performance (makes it 5 times faster).

PR #681, opened 2021-06-17
Truncation position for pane titles

qsmodo:
`r` and `l` flags were added to `tuioptions` to define a same truncation side (`r`ight or `l`eft) for both the active and the inactive panes. The default behavior has not been changed.

Please let me know if there's something to improve. Cheers!


xaizek:
Thanks! I've only made sure that both `r` and `l` don't appear in `'tuioptions'` value at the same time.

PR #686, opened 2021-06-27
Minore additions and fixes for archivemount filetypes

PRESFIL:
I added zstd support for `archivemount` and zstd preview, fixed preview for `*.tar.xz,*.txz` archives.

What do you think? If you need to change something, the branch is editable.


xaizek:
Thanks.

PR #691, opened 2021-07-13
Make :compare honor filters

### LINKED ISSUE #683: Compare directories, ignore filtered (or binary) files?
Suppose one compares two directories with different versions of a C project to find out which `.[ch]` files are different. If the code has been compiled, the compare view becomes much less usable because Vifm will also (correctly) mark the binary files (e.g. `.o`) as different.

Hence an option to disable comparison of binary files might be very handy. Or maybe make the comparison heed filters such as `:filter !/.*\.o/`?

qsmodo:
I think there is no need to update documentation this time since it already says

> Filtered files are not checked in / search or :commands.

Closes #683.

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -42,6 +42,7 @@
 #include "utils/trie.h"
 #include "utils/utils.h"
 #include "filelist.h"
+#include "filtering.h"
 #include "fops_cpmv.h"
 #include "fops_misc.h"
 #include "running.h"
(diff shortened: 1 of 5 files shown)


xaizek:
Thanks! There were a couple of small bugs, see my commits.

qsmodo:
Whoops, sorry for those. About ee1bba9, I did intend to avoid code duplication but on second thought I agree it is not worth the performance penalty.

Couldn't we also move the `full_path` calculation down to only after the dot-files test?

xaizek:
> Couldn't we also move the full_path calculation down to only after the dot-files test?

Right, thanks.

PR #798, opened 2022-06-11
Add hborder

### LINKED ISSUE #712: hborder: Horizontal separator for horizontal split
Vifm uses `vborder` to fill the borders of the window and also to separate the left and right views on a vertical split.

In my opinion it would make sense to also separate the bottom and top views on horizontal splits with a `hborder` character, useful for those who don't color the background of the top-line, such that it is not always obvious where one view ends and the other begins:

qsmodo:
This is meant to close #712.

If you like it, I will document it.

Besides hborder, the main changes are:

1. `vborder=""` is treated equally to `vborder =" "` in `clear_border`.
2. No longer "hide" vborder from `set fillchars?` when it is null or a space. I think doing otherwise would unnecessarily complicate the code, but let me know if it's shortsight.

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -4051,11 +4051,16 @@ Sets characters used to fill borders.
 
   item         default    used for
   vborder:c    ' '        left, middle and right vertical borders
+  hborder:c    ''         middle horizontal border
 
-If value is omitted, its default value is used.  Example:
+A null string for vborder is equivalent to a space.
+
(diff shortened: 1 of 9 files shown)


xaizek:
This looks really good, please add documentation.

qsmodo:
Thank you, Xaizek! There you go.

xaizek:
Separate commits are fine, thanks!

PR #800, opened 2022-06-29
add the curses supported shift key definitions so they can be mapped

mckellygit:
Thank you so much for vifm.

Another method could be to instead call:
    keyok(KEY_SF, FALSE);
etc. for each of these curses defined shifted keys in setup_ncurses_interface() in ui/ui.c
But that would then mean mappings don't use the name such as \<s-down\> but the actual multi-char escape codes.  Not sure which method seems better ?
We could also wrap each in an ifdef to make sure they are defined before using, if needed.

### DIFF
--- a/src/bracket_notation.c
+++ b/src/bracket_notation.c
@@ -504,6 +504,16 @@ static key_pair_t key_pairs[] = {
 	{ L"<insert>",   { K(KEY_IC) }        },
 	{ L"<pageup>",   { K(KEY_PPAGE) }     },
 	{ L"<pagedown>", { K(KEY_NPAGE) }     },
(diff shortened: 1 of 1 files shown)


mckellygit:
@xaizek what do you think ?

xaizek:
Either I didn't notice these before, they are new, or they aren't always present, will need to test. Using `KEY_*` macros is preferred, they will match terminfo entries.

mckellygit:
ok, I probably then also need to add these 10 to wchar_to_spec() ?

mckellygit:
Pushed an update.

xaizek:
Thanks, merged it.

PR #820, opened 2022-09-10
Add shell completion for fish-shell

FollieHiyuki:
Completion for '+' argument is not implemented as I can't figure out how to do so. Help is welcomed.

### DIFF
--- a/data/shell-completion/fish/vifm.fish
+++ b/data/shell-completion/fish/vifm.fish
@@ -1,0 +1,29 @@
+# Only 1 hyphen is allowed
+function __fish_complete_hyphen
+    set -l cmd (commandline -opc)
+    set -e cmd[1]
+    for i in $cmd
+        if test "$i" = "-"
+            return
+        end
+    end
+    printf "-"
+end
+
+complete -c vifm    -F                            -a "(__fish_complete_hyphen)"
+complete -c vifm -r -F        -l "select"                                               -d "Open parent directory of the given path and select specified file in it"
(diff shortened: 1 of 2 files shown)


FollieHiyuki:
I got completion for '+' working, thanks to some advice in the linked issue.

Some drawbacks:
- The first time the user invokes completion (by typing `vifm <TAB>`) it is pretty slow (I'm looping through all available command strings and their descriptions).
- If the user types `vifm +<TAB>` as the first time the completion is invoked, it will only complete commands with '+' element in them (g++, clang++, ...). Every completion afterwards will be broken (as completion for '+' is now cached).

xaizek:
Thanks for this. I think you misunderstood `-c` and `+`. They accept Vifm's command-line :commands, not external commands, so you don't really need to complete thier arguments.

FollieHiyuki:
Silly me :)

I don't use `-c` and `+` personally so I have no idea how they work. The description `run <command> on startup` is kind of misleading for me.

On that note I think it will be nice having a way for `vifm` to print out its available commands (maybe with `vifm --available-commands` or something) so we can fetch it to the completion?

xaizek:
Thanks, merged.

> On that note I think it will be nice having a way for `vifm` to print out its available commands (maybe with `vifm --available-commands` or something) so we can fetch it to the completion script.

`--list-commands` (for consistency with `--list-servers`) can be added, but there is a reason why you don't know how to complete `+c`, it's useful for scripts rather than interactively.

PR #826, opened 2022-09-28
Update fill_side_by_side

IAmKapuze:
This commit solves several problems:
1. Huge memory consumption, that critical for huge comparisons (this method no need extra memory). My example comparing (Linux kernel source files) contains:
curr.nentries: 62132, other.nentries: 56871
Total memory consumption:
62133 (curr.nentries + 1) * 56872 (other.nentries + 1) * 4 (size of int) = 14134511904 bytes ~ 13 GB
\+
62133 (curr.nentries + 1) * 56872 (other.nentries + 1) * 1 (size of char) = 3533627976 bytes ~ 3GB
\=
17668139880 ~ 16.4 GB
I have no this value of memory, and vifm falls (video):

https://user-images.githubusercontent.com/16070551/192818194-64e902a2-8235-4913-8c37-8e3403555a0c.mp4

2. Bad detection of identical files. In my opinion, files with different names shouldn't be at the same line:
image

3. Issues with comparing wrong files:
image
image

### DIFF
--- a/src/compare.c
+++ b/src/compare.c
@@ -311,73 +311,114 @@ fill_side_by_side(entries_t curr, entries_t other, int group_paths)
 	enum { UP, LEFT, DIAG };
 
 	int i, j;
-	/* Describes results of solving sub-problems. */
-	int (*d)[other.nentries + 1] =
-		reallocarray(NULL, curr.nentries + 1, sizeof(*d));
-	/* Describes paths (backtracking handles ambiguity badly). */
-	char (*p)[other.nentries + 1] =
-		reallocarray(NULL, curr.nentries + 1, sizeof(*p));
 
-	for(i = 0; i <= curr.nentries; ++i)
+	if (group_paths)
 	{
-		for(j = 0; j <= other.nentries; ++j)
+		i = 0;
+		j = 0;
+
+		while ((i < curr.nentries) || (j < other.nentries)) {
+			int pathcmp, namecmp;
+
+			if ((i < curr.nentries) && (j < other.nentries)) {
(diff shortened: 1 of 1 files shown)


xaizek:
Actually only 3 GiB are needed... I probably left whole `d` table instead of two of its rows by mistake. Even then something like an algorithm from "An O(NP) Sequence Comparison Algorithm" by Sun Wu, Udi Manber and Gene Myers should do much better for large numbers of files.

The reason your implementation doesn't need extra memory is because it does a different thing: aligns by path instead of computing minimal edit script for a given comparison criterion ("Before" picture has 7 changes vs. 13 in "After"). Which might actually be better when `groupbypath` is used, I might have just retrofitted path grouping into existing code, don't remember. But current implementation is still useful for finding duplicates that have changed their names, which isn't possible with pure path alignment, so it shouldn't be thrown away.

IAmKapuze:
Then should I make my implementation as mode for ***fill_side_by_side***, when ***group_paths == 1***? Or add a new option for compare command?
If first, then what to do with using group_paths by [original method](https://github.com/vifm/vifm/blob/dd8467c452140b63a11e38b96ecfe46fcb1400a1/src/compare.c#L344)?
Can it just be dropped in this case?
```
if (centry->id == oentry->id && d[i - 1][j - 1] <= d[i][j])
```

IAmKapuze:
I updated code at this branch. Could you please comment this?

xaizek:
> Or add a new option for compare command?

No, this path grouping is probably what most people expect to get (e.g., I don't know why only names were compared before, that might not be intentional and I meant to change it). If there will be shortcomings, can tweak `:compare`.

> Can it just be dropped in this case?

Yes, just as if it's `0`.

Thanks for the PR and update, I'll take a closer look and fix tests later.

PR #833, opened 2022-10-20
Added viewing of comparison results

IAmKapuze:
I added simple viewing of comparison results to statusbar, but after any changing of state (for example, moving cursor) this message is disappearance.
I see, that it depend on modes, so should I add new mode for it (for example, add _COMPARE_MODE_, check it into _modes_statusbar_update_ and call own _print_compare_msg_ func) or it is a bad way?

I would like to see comparing results at any time, when statusbar is empty now (no selected files, no VISUAL mode, etc.)

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -6493,6 +6493,9 @@ data.
 Comparison views have second column displaying id of the file, files with same
 id are considered to be equal.  The view columns configuration is predefined.
 
+The status bar displays only the initial result of the comparison and can be
+out of date.
+
 .B Behaviour
 
 When two views are being compared against each other the following changes to
--- a/data/vim/doc/app/vifm-app.txt
+++ b/data/vim/doc/app/vifm-app.txt
@@ -5497,6 +5497,9 @@ data.
 Comparison views have second column displaying id of the file, files with same
 id are considered to be equal.  The view columns configuration is predefined.
 
+The status bar displays only the initial result of the comparison and can be
+out of date.
+
 Behaviour~
 
 When two views are being compared against each other the following changes to
--- a/src/compare.c
+++ b/src/compare.c
@@ -173,6 +173,9 @@ compare_two_panes(CompareType ct, ListType lt, int flags)
 	flist_custom_start(curr_view, lt == LT_ALL ? "diff" : "dups diff");
 	flist_custom_start(other_view, lt == LT_ALL ? "diff" : "dups diff");
 
+	/* Clear previous comparison results for curr_view */
+	memset(&curr_view->custom.diff_cmp_result, 0, sizeof(struct compare_result_t));
+
 	if(group_paths)
 	{
 		fill_side_by_side_by_paths(curr, other, flags);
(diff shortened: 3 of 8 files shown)


xaizek (commented):
You can just check for `cv_compare(view->custom.type)` in `modes_statusbar_update()` and print the message there. For this to work `compare_result_t` needs to be stored in `struct cv_data_t` (same data of both views, I guess).

There is however an issue of this information becoming outdated as lists of files are updated externally or as a result of operations.

IAmKapuze:
> You can just check for `cv_compare(view->custom.type)` in `modes_statusbar_update()` and print the message there. For this to work `compare_result_t` needs to be stored in `struct cv_data_t` (same data for both views, I guess).

Thank you!

> There is however an issue of this information becoming outdated as lists of files are updated externally or as a result of operations.

If internal operations can to be tracked, external updates can't by simple way. May be user just needs to be warned about it?

xaizek:
> If internal operations can to be tracked, external updates can't by simple way. May be user just needs to be warned about it?

I guess if message emphasizes that the stats are given for the initial state, that can work.

xaizek:
> Message like _Initial state of comparison_ is so long for statusbar. Do you agree?
> 
> Can it just be added to the **Hint** (`tips` array)? "_The result of the :compare in the status bar displays only the initial value_"

That's too non-obvious, if information persists I'd expect it to be up-to-date. Can just add `(on compare)` in front to also make it explicit that it's about `:compare` combined with documentation update.

IAmKapuze:
> add `(on compare)` in front

Sorry, what do you mean? Statusbar?

xaizek:
Yes, shorter prefix for the statusbar message.

IAmKapuze:
I added messages to statusbar and help.
Also added `print_compare_msg` to `ui` -> `update_start` for printing message on startup by command:
`vifm Dir1 Dir2 -c compare`.
Can you verify this changes, please?
If all is ok, I will squash commits and undraft this PR

xaizek:
Sorry for the delay.

Another place for printing the message is in https://github.com/vifm/vifm/blob/master/src/event_loop.c#L288, otherwise it disappears when you press <kbd>Ctrl+W</kbd> or alike. Also need to account for swapping panes (<kbd>Ctrl+W</kbd> <kbd>x</kbd>). Pushed corrections to the branch.

IAmKapuze:
> Another place for printing the message is in https://github.com/vifm/vifm/blob/master/src/event_loop.c#L288, otherwise it disappears when you press Ctrl+W or alike. Also need to account for swapping panes (Ctrl+W x). Pushed corrections to the branch.

Thank you!

I saw that the `event_loop` also performs some actions with the statusbar, but understood what exactly only after your message. I will keep that in mind

I squashed commits and undraft PR

xaizek:
Thanks.  Was merged in 1f13438c47d240bfa1acc7379be8435934a1f020.  I amended top commit, so GitHub didn't recognized the merge.

PR #837, opened 2022-11-10
adds mouse support

haolian9:
hi, i'm just trying to make mouse work in vifm. (i have never wrote C before, so i open this pr just want to put out some points, and hope them can help eventually)

currently there are two problems block my way:
1. i found no mouse support is not enabled
2. i failed to find a proper place to add handler for mouse events

for problem 1, i managed to know that:
* vifm uses ncursesw for TUI, so i added the following codes according to [this gist](https://gist.github.com/sylt/93d3f7b77e7f3a881603)
```
keypad(lwin.win, TRUE);
keypad(rwin.win, TRUE);
mousemask(ALL_MOUSE_EVENTS|REPORT_MOUSE_POSITION, NULL);
```
* there are CSI code to enable/control mouse events, so i added these codes according to [this doc](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html)
```
printf("\x1B[?1003h");
```

then i think the problem 1 is solved more or less, because when i use mouse scroll in vifm inside tmux, the tmux's scroll mode won't be triggered anymore, and click on left or right window of vifm i can see some chars printed out in the status line. in this screenshot, the code is `3m`, it's correct because i clicked `src/` line.
9616ef1e-60be-11ed-8a6a-54b203702c9c.

about the problem 2, i have not figured out a clear way to do. vifm has different mode, and each mode has its compile time known `built_cmds`, this goes agains to mouse event, because mouse event consits of several parts, each part can be changed dynamically, here are some events i triggered randomly: 

```
\x1b[<0;49;8M
\x1b[<0;49;8m
\x1b[<2;23;8M
\x1b[<2;23;8m
\x1b[<0;21;1M
\x1b[<0;21;1m
\x1b[<0;3;1M
\x1b[<0;3;1m
\x1b[<65;36;11M
\x1b[<64;36;11M
```
(the pattern is `\x1b[<btn;col;row;press`)

would you give some hints on solving this? it does not seem to be easy change i can handle.


haolian9:
yay! i managed to make right click, scroll up and scroll down work in the latest commit

xaizek:
Hi.

Sorry for the delay. I've quickly rebased (5th or 6th time since 2011..) my branch for mouse support, it's [here](https://github.com/vifm/vifm/compare/latest-mouse-support). Doesn't seem to work as is for some reason, but shows how to handle mouse in different modes. I think handling of scrolling was the only thing that actually worked well, other events were tend to be lost or delayed which made use of the mouse just unbearable (except on Windows). Hence it was never merged.

haolian9:
nice to hear from you, it did show me a good example, thank you!

> other events were tend to be lost or delayed

i did see others have met similar problem in the internet and most of them got solved by adding `keypad(win, TRUE)`,  i'll try it out.

(i'm longing for the mouse support for vifm for a long time, i even had crafted a TUI file browser with mouse support using naive CSI code, so i'm rather sure it is possible to get mouse work properly)

haolian9:
I have totally discarded my previous commits, and the latest-mouse-support does show a clear way to handle mouse events.
Since the click event reporting has been disabled, i removed the handling of button1_double_click, and treats "click on a focused item" as cmd_l in normal.c.
Now in normal mode the left and right click, scroll up and down work as expected on my side: in st+tmux, in st, in st+nvim's terminal.

> other events were tend to be lost or delayed

Actually I have not met this problem. and i was wrong on`keypad()` which is not necessary at least in my tests.

xaizek:
Surprisingly, it actually does seem to work now in `xterm` with and without `tmux`. `urxvt` has a weird behaviour of not sending clicks until I move the mouse, but that might be how it works (maybe it's configurable). Sample `vte` terminal also works.

haolian9:
i think i'm using the wrong mode 1003, which should be 1002. and i forgot to clear the mode when vifm exits.
```
Ps = 1 0 0 2  ⇒  Use Cell Motion Mouse Tracking, xterm.  See
    the section [Button-event tracking](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Button-event-tracking).
Ps = 1 0 0 3  ⇒  Use All Motion Mouse Tracking, xterm.  See
    the section [Any-event tracking](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Any-event-tracking).
```

after changed to 1002, st, urxvt, xterm, kitty, guake, gnome-terminal work fine for left and right click, scroll up and down. no more motion event will be reported.

xaizek:
> i think i'm using the wrong mode 1003, which should be 1002.

Yes, I confirm that it fixes `urxvt`.

> and i forgot to clear the mode when vifm exits.

That might not be necessary. `endwin()` should restore previous state of the terminal.

> yet i can see flickers while scrolling in xterm, urxvt, kitty

It might depend on whether double buffering is enabled or not. Scrolling updates so much of the screen, curses likely redraws it all and without double buffering that will be visible.

haolian9:
Thanks for your attention and the hints! I think my initial goal, to find out whether mouse support is addable to vifm or not, was accomplished. I'll try to solve the cleanup and flicker problem.
Comparison to [vim's mouse support](https://vimhelp.org/term.txt.html#mouse-using), the next step might be supporting mouse relevant kemaps which is beyond my ability for now.

Do you think it's worth to make this pr mergeable with some necessary changes?

xaizek:
> I'll try to solve the cleanup and flicker problem.

There might be nothing to solve there. If flicker is due to lack of double buffering, it's basically unavoidable.

> Do you think it's worth to make this pr mergeable with some necessary changes?

Sure, would be nice if it didn't fail CI on Linux/OS X (seems to happen for 32-bit systems due to curses limitations due to which Button5 isn't accessible). I can handle Windows where API seems to have been updated in newer PDCurses.

***

The rest depends on how far you want to go with it. Proper mouse support should involve:
 * `'mouse'` option to keep this optional
 * `<*Mouse>` and `<*Release>` keys
 * documentation on what mouse buttons do in different modes

Looking at Vim, clicks probably aren't needed, so `mouseinterval(0)` can stay.

Mouse (de)initialization should be done when `'mouse'` changes after it's introduced.

Mouse keys probably need introduction of custom keys (like `WC_C_SPACE`) and mapping events onto them, then modes would handle something like `WC_MOUSE_LPRESS`, etc. instead of generic `KEY_MOUSE`.  This would actually look more like your initial take on this.

Feel free to do as much as you wish (I can give more details if you'll say what you want to do), I'll do the rest.

haolian9:
Thanks for your detailed guides, I'll try to solve the CI failures first, and remove the unnecessary cleanup logic i added before.

haolian9:
Hi xaizek, I have fixed the CI failure, and I successfully built and rant vifm in ubuntu bionic.
Though I still have some questions about current implementation which are expecting your reply.

> modes would handle something like WC_MOUSE_LPRESS, etc. instead of generic KEY_MOUSE
this is quite interesting to me, I'll try it out nextly.

xaizek:
Hi. I see, thanks. So this isn't WIP anymore?

haolian9:
I rebased commits into one, I think this pr is finished for providing basic mouse support.

xaizek:
OK, thanks.

I rebased the branch, made some fixes, changes, updated documentation. Added `'mouse'` option.

> > modes would handle something like WC_MOUSE_LPRESS, etc. instead of generic KEY_MOUSE
> 
> this is quite interesting to me, I'll try it out nextly.

This will be harder than I thought. Just passing different key numbers won't be enough as it won't preserve coordinates. Probably need a queue of mouse events to store information obtained from curses and pop it in handlers.

Other things which need to be addressed in the future:
 * handling of miller/ls-like views
 * supporting more modes, like "view" and "more" modes

Just waiting for CI on this before merging.

haolian9:
Thanks, you completed me :).

Issue #832, opened 2022-10-18
View comparing results (Identical, Different, Unique left, Unique right)


IAmKapuze:
Information about comparing results (Identical, Different, Unique left, Unique right) can be easy got from _fill_side_by_side_ funcs. It can be printed, for example, in _status_bar_.
I would like to help with implementation, but after returning from comparing func _status_bar_ is reset (I tried using _ui_sb_msg_ func at the end of _compare_two_panes_), and I can't understand how to set status bar message to comparing window.

Also, may be there is a way to show only files from single list (Identical/Different/Unique left/Unique right) without reexec comparing (for example, make additional separated list (views) for each result type and switch between them with commands (_showall_, _showidentical_, _showdifferent_, _showuniqueleft_, _showuniqueright_))?

xaizek:
> Information about comparing results (Identical, Different, Unique left, Unique right) can be easy got from _fill_side_by_side_ funcs. It can be printed, for example, in _status_bar_.

You mean printing statistics? You need to return non-zero from `compare_two_panes()` to preserve status bar message.

> Also, may be there is a way to show only files from single list (Identical/Different/Unique left/Unique right) without reexec comparing (for example, make additional separated list (views) for each result type and switch between them with commands (_showall_, _showidentical_, _showdifferent_, _showuniqueleft_, _showuniqueright_))?

That would require always doing the full diff and preserving it to process later. Current implementation throws averything away when it's done and might not even build the full state at any point if it wasn't needed to execute the command (don't know if it saves much resources though).

IAmKapuze:
> You mean printing statistics? You need to return non-zero from `compare_two_panes()` to preserve status bar message.

Yes, it works. Thanks
I did some draft implementation, but I have some difficulties (questions). I will create draft PR

Also I think about switching between result lists. For best user-experience it should be toggled (for example, only Unique Left/Unique Right lists can be displayed). What do you think about it?

xaizek:
> Also I think about switching between result lists. For best user-experience it should be toggled (for example, only Unique Left/Unique Right lists can be displayed). What do you think about it?

This probably needs a set of `show*` arguments for `:compare` to specify initial state or switch to the desired one of the last comparison.

IAmKapuze:
May be diff can be stored fully, while comparison window is open? It can be threw after closing comparison

xaizek:
`:compare listunique` doesn't create a compare view at all, so that would take a bit more tracking to find out when to free comparison results.

[...]

PR #839, opened 2022-11-12
Comparison results selecting

IAmKapuze:
Added toggled `showidentical | showdifferent | showuniqueleft | showuniqueright` arguments for :compare command. If current view is `cv_compare` then arguments will be toggled with existed before comparing.

Now applying of them is restarting comparing. I don't know which is better. May be result of `flist_custom_put` calls can be saved to self-erased (on exit from compare view) structure, but I doubt.

Also disabled printing stats for `ofone` mode by moving `single_pane` to `CompareFlags`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -14,7 +14,8 @@
 	columns.  Thanks to Aleksandr Vysotskiy (a.k.a. loki1368).
 
 	Added show* arguments to :compare command to control/switch which sets of
-	files are displayed.  Patch by Alexandr Keyp (a.k.a. IAmKapuze).
+	files are displayed (toggling is done by :compare!).  Patch by Alexandr
+	Keyp (a.k.a. IAmKapuze).
 
 	Reduced amount of memory consumed by `:compare groupids`.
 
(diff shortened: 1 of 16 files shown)


xaizek:
Thanks. Looks good overall.

> Now applying of them is restarting comparing. I don't know which is better. May be result of `flist_custom_put` calls can be saved to self-erased (on exit from compare view) structure, but I doubt.

Caching results would be obviously better for performance, but it's more complicated and error prone to do (need to somehow take into account changes performed on the lists). This implementation doesn't make anything slower, so I think it's fine as is and can be improved in the future if needed.

IAmKapuze:
Then, can we just rename issue #832 to "cache comparison results" and keep its open as [improvement/addition](https://github.com/vifm/vifm/labels/improvement%2Faddition) for future, after completing(merging) this task?

xaizek:
I think we can just close the issue when this is merged and deal with the caching separately when that comes up.

xaizek:
Made some changes to functionality:
 * toggling is done via `:compare!`
 * toggling preserves original comparison options rather than using defaults

PR #809, opened 2022-08-06
vifmrc modernization

dmlemos:
Mostly changes to `vifmrc-osx`, with some common options cross polinated.

## Reasons for changing some of the options
* MPlayerX is deprecated. IINA seems to be the successor
* Add rar extract files to the osx version. Requires `unrar` tool. Since it a common format I believe it will help new users, as it would for me. Unrar can be installed using homebrew.
* Use `ffprobe` for all fileviewer operations, since it supports pretty much everything. Doesn't require any extra tools
* Add verbose mode to tar arguments for progress. Specially useful for extracting big files, otherwise nothing shows on screen
* `shasum` is the native way on mac. Otherwise it requires `coreutils` installed. Also easier to add new checksums
* Use `tar` to handle zip files, as I discovered it is much faster

Comparison between the two:
```shell
$ time unzip -d EUROPATheLastBattle_unzip EUROPATheLastBattle.zip
Archive:  EUROPATheLastBattle.zip
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 1.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 7.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 6.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 8.mp4
 extracting: EUROPATheLastBattle_unzip/Europa-TheLastBattle-Part5finalCut.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 4.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Trailer   Coming Soon (2017).mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 2.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 3.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 10.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 9.mp4
 extracting: EUROPATheLastBattle_unzip/EUROPA - The Last Battle - Part 5.mp4

real    1m48.206s
user    0m50.857s
sys     0m15.330s

$ mkdir EUROPATheLastBattle_tar
mkdir: created directory 'EUROPATheLastBattle_tar'
[...]

PR #840, opened 2022-11-12
Fix zip fileviewer in vifmrc

IAmKapuze:
I didn't check viewing of `*.jar | *.war | *.ear | *.oxt`, but googling says that it should be works. `*.zip` viewing is works.
If trying view `*.zip` with tar, it will be print `tar: This does not look like a tar archive`


xaizek:
Thanks. Use of `tar` for zip was added in #809. I guess it works on OS X.

PR #861, opened 2023-01-10
Fix incorrect plugin loading path

haolian9:
Hi, with the latest `--plugins-dir` option, I found vifm will always load plugin under `cfg.config_dir` (in my case, it is `$HOME/.config/vifm/plugins`), but it should respect the `plug->real_path`.
So I made this quick fix without looking too deep into the source.

todo:
* is the name param still necessary? since we get full_path from `plug->real_path` now
* add a test

### DIFF
--- a/src/lua/vlua.c
+++ b/src/lua/vlua.c
@@ -23,7 +23,6 @@
 #include <stdlib.h> /* free() */
 #include <string.h> /* strdup() */
 
-#include "../cfg/config.h"
 #include "../cfg/info.h"
 #include "../compat/dtype.h"
 #include "../compat/fs_limits.h"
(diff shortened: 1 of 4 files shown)


haolian9:
wow, there are not a few failed tests need to be fixed.

xaizek:
Hi,

Thank you, I didn't notice the wrong one is being loaded. I removed `name` and fixed existing tests, so CI should pass now.

xaizek:
I've updated my test to verify which plugin is loaded too. It was easier to do then to explain how it can be done. (The new check should have been there from the start.)

haolian9:
Yay, much thanks for your help! Hours ago, I was struggling with compile time string concat for the broken tests, haha.

Issue #855, opened 2022-12-27
feature request - clickable tab line


haolian9:
Hi, this need had found me eventually, haha. Would you consider supporting it?

That's what I got so far:
1. `wenclose(tab_line, e.y, e.x)` in normal.c  to determine whether the user has clicked on the tab line
2. finding a proper way to determine which tab has been clicked
3. `tabs_goto()`

I failed to find a way to make step 2 happen. The easiest way I can imagine is processing the content of the tab line, `[1: foo] [2: bar]`, but I'm afraid it's not a reliable way; Also I'm hijacking mouse events in normal mode, while i think the mouse events for the tab line should be mode-agnostic.

haolian9:
Or maybe I can switch previous and next tabs using mouse scroll events respectively as an alternative, and tabs_previous and tabs_next should make it done more easily.

xaizek:
Hi,

Step 2 is easy to do if you have an array specifying width of each tab label. `print_tabline()` in `src/ui/ui.c` has this information while formatting the tab line, but it's not stored anywhere.

> the mouse events for the tab line should be mode-agnostic.

Vim seems to terminate visual mode when you click a tab, but not a command-line mode. Normal mode is definitely the most common case for handling this.

> Or alternatively I can switch previous and next tabs using mouse scroll events, and tabs_previous and tabs_next are already available for that.

Both ways can be implemented, but this one is simpler.

haolian9:
Thanks for considering this! And the hints!

PR #862, opened 2023-01-10
switch tabs when mouse scroll events happen on tab line in normal mode

haolian9:
Hi, this is a possible and easy way to suffice #855. do you think it is acceptable?

I have used this solution about two weeks, it does its job greatly :)


xaizek:
Hi,

This handles only global tabs, pane tabs are drawn on `curr_view->title` curses window. I think both should be supported.

haolian9:
oh, I had not noticed the difference, will look into it further.

haolian9:
It is &panescope that matters, I had got it wrongly. Thanks, and Sorry you have to finish it yourself in every PR i had ever made.

Issue #819, opened 2022-09-07
Edit Registers with Vim


BioBox:
I believe it would be very useful if there was a way to edit a given register with $EDITOR and save it. Could you add it to `TODO`?

xaizek:
Can you provide a use case? Do you want to load registers and append to them? Just editing can produce useless registers which don't list paths to actual files.

BioBox:
Okay sure. I like to use registers for making playlists of videos, because vifm makes it really easy to grab an arbitrary selection of files.

But wait! I want to remove an entry, or I want to reorder it...
Another problem is that when I move my data to a computer that has different mount points it breaks them all. This could be easily fixed with vim's block line mode if I had the ability to edit regiesters.

xaizek:
Thanks, that sounds like it should be added. Probably need a new command like `:regedit`.

PR #864, opened 2023-01-23
Regedit feature

nullptr-deref:
Implemented command requested in #819.
All registers can be edited except blackhole and yet empty registers.

### DIFF
--- a/src/cmd_handlers.c
+++ b/src/cmd_handlers.c
@@ -114,7 +114,6 @@
 #include "undo.h"
 #include "vifm.h"
 
-static int regedit_cmd(const cmd_info_t *cmd_info);
 static int goto_cmd(const cmd_info_t *cmd_info);
 static int emark_cmd(const cmd_info_t *cmd_info);
 static int alink_cmd(const cmd_info_t *cmd_info);
@@ -263,6 +262,7 @@ static int qmap_cmd(const cmd_info_t *cmd_info);
 static int qnoremap_cmd(const cmd_info_t *cmd_info);
 static int qunmap_cmd(const cmd_info_t *cmd_info);
 static int redraw_cmd(const cmd_info_t *cmd_info);
+static int regedit_cmd(const cmd_info_t *cmd_info);
 static int registers_cmd(const cmd_info_t *cmd_info);
 static int regular_cmd(const cmd_info_t *cmd_info);
 static int rename_cmd(const cmd_info_t *cmd_info);
@@ -1013,57 +1013,6 @@ goto_cmd(const cmd_info_t *cmd_info)
 {
 	cmds_preserve_selection();
 	fpos_set_pos(curr_view, cmd_info->end);
-	return 0;
-}
-
-/* Opens external editor to edit specified register contents. */
-static int
-regedit_cmd(const cmd_info_t *cmd_info)
-{
-	const int regname = cmd_info->argc > 0 ? tolower(cmd_info->argv[0][0]) : DEFAULT_REG_NAME;
-	if(regname == BLACKHOLE_REG_NAME)
-	{
-		ui_sb_err("Cannot modify blackhole register.");
-		return CMDS_ERR_CUSTOM;
-	}
-
-	reg_t *reg = regs_find(regname);
-	if(reg == NULL)
-	{
-		ui_sb_err("Register with given name does not exist.");
-		return CMDS_ERR_CUSTOM;
-	}
-	char tmp_fname[PATH_MAX + 1];
-	generate_tmp_file_name("vifm.regedit", tmp_fname, sizeof(tmp_fname));
-
-	mode_t saved_umask = umask(~0600);
-	const int write_result = write_file_of_lines(tmp_fname, reg->files, reg->nfiles);
-	(void)umask(saved_umask);
-	if(write_result != 0)
-	{
-		ui_sb_err("Couldn't write register content into external file.");
-		return CMDS_ERR_CUSTOM;
-	}
-	// Forking is disabled because in some cases process can return value
-	// before any changes will be written and editor will be closed.
-	const int edit_result = vim_view_file(tmp_fname, 1, 1, /*allow_forking=*/0);
-	if(edit_result != 0)
-	{
-		ui_sb_err("Register content edition went unsuccessful.");
-		return CMDS_ERR_CUSTOM;
-	}
-
-	int read_lines = 0;
-	char **edited_content = read_file_of_lines(tmp_fname, &read_lines);
-	unlink(tmp_fname);
-	if(edited_content == NULL)
-	{
-		ui_sb_err("Couldn't read register editions from the external file.");
-		return CMDS_ERR_CUSTOM;
-	}
-	reg->files = edited_content;
-	reg->nfiles = read_lines;
-
 	return 0;
 }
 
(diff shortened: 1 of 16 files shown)


xaizek:
Thanks, that's basically the implementation. Although, why not edit empty registers? They aren't really special, one might want to load list of files into an empty register.

nullptr-deref:
I was thinking of that, of course, but my ~~laziness took over me~~ I see both pros and cons of implementing access to empty registers.
Basically, if you want to store a bunch of filenames into a register, you'd probably do it by just selecting it via vifm's selection tool and yank them into the register. On the other hand, it's probably a bit more convenient to manually type a list of files contained in different directories which are far apart from each other, but that's an arguable statement, in my opinion.
Either way, I'm going to hold development of it for a couple of days due to my exams and then we'll see.

nullptr-deref:
Another thought that've crossed my mind is that one might want to load a list of filenames from external file right into the register but I can't see any use case of this whatsoever.
Anyway, it's more about expanding functionality of the actual command and I don't think that it's right to overwhelm just one command with a lots of meanings.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
 	return 0;
 }
 
+static int
+regedit_cmd(const cmd_info_t *cmd_info)
+{
+	if (cmd_info->argc > 1)
+		return CMDS_ERR_TRAILING_CHARS;
+
```
xaizek, line 1024:
```suggestion
```

Because of `.min_args = 0,   .max_args = 1`, `argc` will always be either `0` or `1`, so this if-statement is unnecessary.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
 	return 0;
 }
 
+static int
```
xaizek, line 1019: Please add a short comment above like `/* Opens an editor to edit register contents. */`.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+regedit_cmd(const cmd_info_t *cmd_info)
+{
+	if (cmd_info->argc > 1)
+		return CMDS_ERR_TRAILING_CHARS;
+
+	const int regname = cmd_info->argc > 0 ? cmd_info->argv[0][0] : DEFAULT_REG_NAME;
+	if (regname == BLACKHOLE_REG_NAME)
+	{
+		ui_sb_err("Cannot modify blackhole register.");
+		return 1;
```
xaizek, line 1029: Although it's not used consistently at the moment, returning `CMDS_ERR_CUSTOM` in case of error messages is preferred.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+	}
+
+	reg_t *reg = regs_find(tolower(regname));
+	if (NULL == reg || reg->nfiles == 0)
+	{
+		ui_sb_err("Nothing to modify in the empty register.");
```
xaizek, line 1035: Stylistic nit 1: `reg == NULL` (no need for yoda comparisons, compiler will warn on `=` in conditional).
Stylistic nit 2: `if(` (no space).

This error message will probably end up saying `No such register: %c`.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+		return 1;
+	}
+	const int fname_len = PATH_MAX + 1;
+	char *tmp_fname = malloc((fname_len) * sizeof(char));
```
xaizek, line 1039: It's fine to have buffer on the stack.  In case you've changed implementation to make `vim_edit_files()` work, just use `vim_view_file()`.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+	generate_tmp_file_name("vifm.regedit", tmp_fname, fname_len);
+	write_file_of_lines(tmp_fname, reg->files, reg->nfiles);
+	const int edit_result = vim_edit_files(1, &tmp_fname);
```
xaizek, line 1042: Both functions return errors, which should be handled.  Temporary file should be removed on error or after reading it (no need to handle failure to remove it).

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
 	return 0;
 }
 
+static int
+regedit_cmd(const cmd_info_t *cmd_info)
+{
+	if (cmd_info->argc > 1)
+		return CMDS_ERR_TRAILING_CHARS;
+
+	const int regname = cmd_info->argc > 0 ? cmd_info->argv[0][0] : DEFAULT_REG_NAME;
```
xaizek, line 1025: `tolower(cmd_info->argv[0][0])` to not do it later.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,44 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+		return 1;
+	}
+	const int fname_len = PATH_MAX + 1;
+	char *tmp_fname = malloc((fname_len) * sizeof(char));
+	generate_tmp_file_name("vifm.regedit", tmp_fname, fname_len);
+	write_file_of_lines(tmp_fname, reg->files, reg->nfiles);
+	const int edit_result = vim_edit_files(1, &tmp_fname);
+
+	regs_clear(tolower(regname));
+	reg->files = read_file_of_lines(tmp_fname, &reg->nfiles);
```
xaizek, line 1045: This clears register on failure to read the file. Generally replacement of state is done like this:
1. Obtain new state.
2. Verify it.
3. Replace old state with the new one.

xaizek (changes requested):
Don't get scared by the density of comments, many of them are trivial.

You've probably seen `edit_list()`:
https://github.com/vifm/vifm/blob/836c575ebf751ebd00bbf1d61dac1256be5f0df0/src/fops_common.c#L1050-L1108
This command will look similar, including `umask()` for the temporary file.

src/cmd_handlers.c
nullptr-deref: Done in ab1010a.

src/cmd_handlers.c
```diff
@@ -1035,20 +1035,25 @@ regedit_cmd(const cmd_info_t *cmd_info)
...
 	// TODO: implement temporary file deletion
+	// TODO: handle editing error
 
-	if(reg->files == NULL)
+	int read_lines = 0;
+	strlist_t *edited_content = read_file_of_lines(tmp_fname, &read_lines);
+	if (edited_content == NULL)
 	{
-		ui_sb_err("Failed to write editions to register.");
+		// TODO: read errno and modify message according to it.
```
xaizek, line 1051: No need for this.  `errno` isn't reliable at this point because `read_file_of_lines()` doesn't try to preserve it.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,48 @@ goto_cmd(const cmd_info_t *cmd_info)
...
 }
 
+/* Opens external editor to edit specified register contents. */
+static int
+regedit_cmd(const cmd_info_t *cmd_info)
+{
+	const int regname = cmd_info->argc > 0 ? tolower(cmd_info->argv[0][0]) : DEFAULT_REG_NAME;
+	if(regname == BLACKHOLE_REG_NAME)
+	{
+		ui_sb_err("Cannot modify blackhole register.");
```
xaizek, line 1026: You forgot to adjust the error after modifying condition.

xaizek (commented):
Just a couple comments on recent changes.

src/cmd_handlers.c
xaizek: Right, I confused them.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,48 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+		return CMDS_ERR_CUSTOM;
+	}
+
+	reg_t *reg = regs_find(regname);
+	if(reg == NULL)
+	{
+		ui_sb_err("Trying to edit uninitialized register.");
```
xaizek, line 1033: `regs_find()` returns `NULL` for registers that don't exist at all, rather than uninitialized ones, which would imply that they can be somehow initialized and then used.

src/cmd_handlers.c
nullptr-deref: I don't fully understand how I should handle these errors. For now this issues just raise error messages.

nullptr-deref: See e689d14.

xaizek: That's fine. By handling I meant that they shouldn't be ignored and should trigger cleanup and error message.

src/cmd_handlers.c
```diff
@@ -1011,6 +1016,56 @@ goto_cmd(const cmd_info_t *cmd_info)
...
+	const int edit_result = vim_view_file(tmp_fname, 1, 1, /*allow_forking=*/1);
+	if(edit_result != 0)
+	{
+		ui_sb_err("Register content edition went unsuccessful.");
+		return CMDS_ERR_CUSTOM;
+	}
+	(void)umask(saved_umask);
+
+	int read_lines = 0;
+	strlist_t *edited_content = read_file_of_lines(tmp_fname, &read_lines);
```
xaizek, line 1055: The type should be `char **`.  You can run `./configure` with `--enable-developer` or `--enable-werror` to treat warnings as errors, otherwise they are easy to miss.  This however might break the build if your toolchain warns about code that's already in the `master`.

nullptr-deref: > This however might break the build if your toolchain warns about code that's already in the `master`.

And it does.

nullptr-deref: Well, this time it didn't cause any issues with `-werror`.

src/cmd_handlers.c
xaizek: My bad, I missed that `read_file_of_lines()` promises to preserve `errno`.

xaizek (approved):
I rebased and squashed your commits so they don't break `git bisect` and gave a message that follows [these rules](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html). I've also realized that reg syncing and canonicalization of paths is necessary, so that's added along with sorting and deduplication.

Now just waiting for CI.

PR #882, opened 2023-02-28
Fix potential null dereference

ZhipengXue97:
**Description**
Fix the null dereference warning detected by static analyse tool infer@facebook

**Warning Type**
Null Dereference

**Component Name**
vifm/src/utils/trie.c(line: 99)
vifm/src/utils/trie.c(line: 320)

**Additional Information**
pointer `trie` last assigned on line 99 could be null and is dereferenced at line 100.
pointer `trie->nodes[bank]` last assigned on line 320 could be null and is dereferenced at line 323.

### DIFF
--- a/src/utils/trie.c
+++ b/src/utils/trie.c
@@ -97,6 +97,10 @@ trie_t *
 trie_create(trie_free_func free_func)
 {
 	trie_t *trie = calloc(1U, sizeof(*trie));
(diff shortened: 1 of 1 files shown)


xaizek:
Thanks. Also realized another issue with `make_node()` while reviewing the changes.

Issue #918, opened 2023-07-16
`--remote-expr` output gets truncated


rr-:
When I'm in a directory that holds a lot of files and I yank all of them to a register: `:%y z` and then display its content: `:echo expand("%rz")`, this works well, and I can see all the files.

Now, when I do the same using `--remote-expr`:

```console
$ vifm --server-name vifm --remote-expr 'expand("%rz")'
```

the output gets truncated. Trying this with different folders of varying path size and file naming flavors, yields always the same output:

```console
$ vifm --server-name vifm --remote-expr 'expand("%rz")'|wc -c
8155
```

This number, 8155, is not arbitrary, and it comes from the IPC buffer being only capable to hold fixed amount of data, whereas the rest is its payload structure: https://github.com/vifm/vifm/blob/master/src/ipc.c#L627-L629

PR #919, opened 2023-07-16
Fix bigger payloads in IPC

rr-:
A proposal of a lazy fix for #918 that tries to format the package multiple times until the buffer is big enough.


xaizek:
Thanks.

PR #921, opened 2023-07-22
Fix xclip keybinds

Gregaras:
I copied xclip keybinds from [this](https://wiki.vifm.info/index.php/How_to_copy_path_to_current_file/directory_to_system_clipboard) article to my config and realised that two keybinds (path yank ones) are already there but not working. 

Added `-n` argument to `echo` and `-selection clipboard` to `xclip`.


xaizek:
Thanks!

PR #927, opened 2023-08-02
Handle other types of files in ueberzug plugin

mahor1221:
Adds support for video, audio, font, pdf, epub, djvu and comic book formats with proper caching.

Acknowledgment: Many of the thumbnailer commands are copy pasted from [vifmimg](https://github.com/thimc/vifmimg).

Optional dependencies:
* ImageMagick
* [ffmpegthumbnailer](https://github.com/dirkvdb/ffmpegthumbnailer)
* [fontpreview](https://github.com/sdushantha/fontpreview)
* pdftoppm
* [ddjvu](http://djvu.sourceforge.net/doc/man/ddjvu.html)
* ~[epub-thumbnailer](https://github.com/marianosimone/epub-thumbnailer)~ [gnome-epub-thumbnailer](https://gitlab.gnome.org/GNOME/gnome-epub-thumbnailer)
* [fuse-zip](https://bitbucket.org/agalanin/fuse-zip)
* [rar2fs](https://github.com/hasse69/rar2fs)
* [fuse3-p7zip](https://github.com/andrew-grechkin/fuse3-p7zip)
* [archivemount](https://github.com/cybernoid/archivemount)

PR #929, opened 2023-08-22
Fix ueberzug usage example for Gif files

mahor1221:
`#ueberzug#image` can't generate thumbnail for Gif files.

So, I added explanation for two possible choices:
 - Having animated previews by passing the gif file directly to `ueberzugpp` with `#ueberzug#image_no_cache`
 - Or having the usual previews with `ueberzug` which uses the `#ueberzug#video` function instead.

### DIFF
--- a/data/plugins/ueberzug/init.lua
+++ b/data/plugins/ueberzug/init.lua
@@ -11,10 +11,19 @@ Ueberzug:
  - improved rewrite in C++: https://github.com/jstkdng/ueberzugpp
 
 Usage example:
-    fileviewer <video/*>,{*.gif}
+    " If you are using ueberzugpp, you can have animated gif previews with this:
+    fileviewer {*.gif}
+        \ #ueberzug#image_no_cache %px %py %pw %ph
+        \ %pc
+        \ #ueberzug#clear
+
+    " Otherwise, use the video fileviewer for gifs
+    " fileviewer <video/*>,{*.gif}
+    fileviewer <video/*>
         \ #ueberzug#video %px %py %pw %ph
         \ %pc
         \ #ueberzug#clear
+
     fileviewer <image/*>
         \ #ueberzug#image %px %py %pw %ph
         " or \ #ueberzug#image_no_cache %px %py %pw %ph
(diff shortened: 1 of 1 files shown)


xaizek:
Hm, I just noticed that GIF preview isn't anymated anymore.  I don't remember whether `ueberzug` could do it, but `ueberzugpp` can definitely show an animation.  Will need to return animated preview while merging this.

mahor1221:
Oh, I didn't know that. I'll update this PR.

PRESFIL:
> Hm, I just noticed that GIF preview isn't anymated anymore.

@xaizek, This is most likely due to the fact that the lua plugin was written
before gif support was removed. There was [a moment][1] when @thimc first disabled
gif support by default, and then completely [abandoned][2] it. But I consider that
this is not a big loss for me.

[1]: https://github.com/thimc/vifmimg/commit/96ef3a4beff485ab697232075739de5c035bcfa6
[2]: https://github.com/thimc/vifmimg/commit/3372889d5bfd65cb590e2e3411813ec98676a392

thimc:
> There was [a moment](https://github.com/thimc/vifmimg/commit/96ef3a4beff485ab697232075739de5c035bcfa6) when @thimc first disabled gif support by default, and then completely [abandoned](https://github.com/thimc/vifmimg/commit/3372889d5bfd65cb590e2e3411813ec98676a392) it. But I consider that this is not a big loss for me.

Just for clarification, the "gif" support you are refering to was awful. It was buggy as heck and I haven't looked in to it ever since. And I decided that in the end, I couldn't care less for it because I don't tend to use GIF files all that often so didn't want something half finished merged in to upstream. I was and am still willing to accept a PR for this in case there's demand for it of course.

xaizek:
@PRESFIL Looks like original `ueberzug` just never supported GIFs.  `uberzugpp` plays GIFs on its own, you just pass it a file name and watch animation in preview without any loops.  #927 made that stop because it attempted to generate GIF previews even when `ueberzugpp` was in use.

xaizek:
Thanks!

PRESFIL:
> `uberzugpp` plays GIFs on its own, you just pass it a file name and watch animation in preview without any loops
Thanks!

PR #931, opened 2023-08-24
saf 0.16 integration, Vifm plugin

dusanx:
Hi,
[saf](https://github.com/dusanx/saf) author here.

> saf is a simple, reliable, rsync-based, battle tested, well rounded backup system, written in Python.
>
> It uses rsync to incrementally back up your data to a different directory, hard disk or remote server via SSH. All operations are incremental, atomic and optionally possible to resume.
>
> saf is using very limited set of commands in the background, so it plays well with storage shells with limited set of commands, like Hetzner Storage Box.
>
> Main backup code is used in production for many years and has proved its reliability, while recent saf updates only improve user facing commands. No guarantees but should be safe to use.

Since saf provides time-machine like backup, being Vifm user for years, I wrote plugin that uses saf to provide real time-machine experience in Vifm. If you are using saf and in Vifm looking at the dir that's under saf backup, `:SafRevisionsToggle`, `:SafPreviousRevision`, and  `:SafNextRevision` are all that's needed to find previous file/folder version in a backups, compare or copy to the original location. Very useful and offers functionality that I wanted to reach when making saf.

I'll keep maintaining plugin unless somebody more capable steps in. Thank you.

### DIFF
--- a/data/plugins/saf-revisions/README.md
+++ b/data/plugins/saf-revisions/README.md
@@ -1,0 +1,21 @@
+# saf-revisions Vifm plugin
+
+Install vifm plugin by copying or symlinking saf-revisions (plugin folder) to $VIFM/plugins/ (usually ~/.config/vifm/plugins/ or ~/.vifm/plugins/). This will make :Saf* commands available in Vifm:
+
+```
+:SafRevisionsToggle     -- Activate or deactivate saf-revisions
+:SafRevisionsActivate   -- Turn saf-revisions on
+:SafRevisionsDeactivate -- Turn saf-revisions off
+:SafPreviousRevision    -- Starfield! Warp! Magic! Back in time.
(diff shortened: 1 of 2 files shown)


xaizek:
Hello,

thanks for the plugin.  `saf` looks interesting (although `.saf.conf` laying in various places might be a bit annoying, but maybe it's not in practice).

You might want to mention somewhere that Python 3.10 is required.  Slackware 15.0 has 3.9 and I had to figure out which version added `match`.

dusanx:
Thanks for merging. Also for the version tip, I'll add that to documentation.

.saf.conf was picked few years ago as a convenient way, somewhat inspired by .git/config, to stick backup source 'head' whenever needed. I even have several nested .saf.conf locations so one backup picks up all and another some folders. Of course may or may not be suitable for different users.

Thanks for the Vifm, I am dedicated user for years, so it only made sense to cover my needs first. Great file manager.

Issue #546, opened 2020-05-02
Using %a in statusline to show the free space gets a wrong value


yanzhang0219:
Hi,

Based on the doc [here](https://vifm.info/vimdoc.shtml#vifm-%27statusline%27), I want to use `%a` in the statusline to show the amount of free space, but I get a weird number `72 T` there. My SSD is 512 so this 72 thing is weird. 

My config about statusline:
![image](https://user-images.githubusercontent.com/11582667/80857826-df780300-8c09-11ea-9076-afab35d28e71.png)

The statusline result:
![image](https://user-images.githubusercontent.com/11582667/80857813-ce2ef680-8c09-11ea-9e84-e91b5e82bb74.png)

Thank you~

xaizek:
Hello,

thanks for the bug report. I didn't realize that on OS X (which you mentioned in another issue) API for querying file-system information has different meaning of fields compared to other systems, which is why you get wrong results.

yanzhang0219:
Oh yeah, sorry I forgot to put my OS version there. Thank you.

yanzhang0219:
Could you please fix it on the master so that I can rebuild it? Thanks!!

yanzhang0219:
Hi,

Thank you for fixing it. And one more issue comes up...

After I run git pull the latest commit from the master and rebuilt it, and when I tried to open vifm, an error popped up:
![image](https://user-images.githubusercontent.com/11582667/80998777-625eb080-8df8-11ea-9512-f1d9f9ee192b.png)

The icons file is [here](https://github.com/yanzhang0219/dotfiles/blob/master/.config/vifm/icons). And if in the icons file I appended an empty line at the end, this error was fixed. 
![image](https://user-images.githubusercontent.com/11582667/80999037-d39e6380-8df8-11ea-8e9d-495c37ceb29c.png)

voh9eepah:
It seems that same fix is needed also for FreeBSD.
[STATVFS(3)](https://man.freebsd.org/cgi/man.cgi?query=statvfs&sektion=3&n=1)

PR #936, opened 2023-09-10
Fix get_drive_info on FreeBSD

voh9eepah:
FreeBSD needs same fix as OS X. See

https://man.freebsd.org/cgi/man.cgi?query=statvfs&sektion=3&n=1

### DIFF
--- a/src/utils/utils_nix.c
+++ b/src/utils/utils_nix.c
@@ -1245,8 +1245,8 @@ get_drive_info(const char at[], uint64_t *total_bytes, uint64_t *free_bytes)
 		return -1;
 	}
 
-#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


xaizek:
Thank you.  Saw your  https://github.com/vifm/vifm/issues/546#issuecomment-1712452793 yesterday, but wanted to check things more closely. Linux turned out to be the odd one, other systems (or at least majority of them) behave like OS X/Free BSD. Other BSDs:
 - https://man.openbsd.org/statvfs
 - https://man.netbsd.org/statvfs.5
 - https://man.bsd.lv/DragonFly-5.6.1/man2/statvfs.2

Plus the code was a bit wrong for Linux as well.

PR #939, opened 2023-09-15
Fix Ueberzug gif example

mahor1221:
Sometimes the file is actually `mp4`, but the extension is `gif`, which will cause the previewer to not work correctly. It's better not to rely on extensions when possible.


xaizek:
Thanks.

PR #945, opened 2023-10-03
Cut the middle part of a long filename

### LINKED ISSUE #868: [Feature] Omit/cut the middle part of a long filename
I've been using YouTube downloaders and now I have lots of files with very long names, with descriptions, subtitles and lyrics in different languages.

I've tried using `{fileext}`, but while there is important information at the beginning of the name, the end before the last extension often contains additional extensions, like language tag on a txt or subtitle file, which are hidden in long names and can become a confusing mess in short names (though it's partly because of #867):
```
x         .desc…
x         .desk…
x.en      .txt
x.info    .json
x         .jpg
x.nordlic….jpg
x.ru      .txt
x.ru.txt  .bak
x         .txt
x.txt     .bak
x         .webm
```

So I thought of the following, which is what mc does:

it does not touch the short names, and cuts the middle out of the long names to fit them:

Screenshot_2023-01-30_19-02-00

If it's not resource-intensive (*maybe this and even some other features could be disabled optionally for directories with too many files?*), I would have something like this, a customizable amount of characters or percentage from the dot:

shorter_name.txt
very_very_extremely_overlong_file_name_that_does_not_fit.en.txt
⇓
shorter_name.txt
very_very_extremely_overlong_file_name_…not_fit.en.txt

VadimCurca:
I often find myself working with files that have long names and for which the beginning and end part is important. I implemented this as an alignment option, which is like left alignment but cuts off the middle part if the name is too long:

image

Closes #868

### DIFF
--- a/data/man/vifm.1
+++ b/data/man/vifm.1
@@ -7332,6 +7332,15 @@ right align (default)
   set viewcolumns={name}
 .EE
 .IP \- 2
+middle align
+
+It's like left alignment, but when the text is bigger than the column, it
+is cut in the middle (so the start and the end of the field are always visible).
+
+.EX
+  set viewcolumns=^{name}
+.EE
+.IP \- 2
 dynamic align
 
 It's like left alignment, but when the text is bigger than the column, the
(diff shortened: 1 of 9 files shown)


xaizek:
One issue with this: search match highlighting can be wrong if match is after ellipsis.  I just wrote it down to `BUGS` because it's not the only issue with that highlighting and they might as well be addressed all at the same time (highlighting code just doesn't have information about what was cut off).

VadimCurca:
Thank you!
Indeed, I didn't notice that. I will see if I can propose a fix for this.

PR #952, opened 2023-10-29
Fix search match highlighting issues with handling lines with hidden parts

VadimCurca:
I have moved the code for adjusting the search match ranges closer to where the line is cut and decorated. I fixed the issues I noticed at the beginning, but if there are still issues I didn't notice or you have change suggestions, let me know.


xaizek:
Thank you for fixing it!  Doing this in two separate places was a bad idea indeed.  I've moved part of the code back to avoid `column_view` unit having to know anything about `fileview` and avoid duplication.  Will merge after doing more testing.

PR #960, opened 2023-11-23
Add zstd and tzst to Default-256.vifm

aleksejrs:
I just added them in the form that I am seeing in the GitHub editor.


xaizek:
Thanks.

PR #970, opened 2023-12-17
Fix little grammar error in documentation

Mrestof:
My first contribution to open source. Wanna help, even if it's not much.


xaizek:
Thanks!  The problem with small things like typos is that it's easy to not bother fixing or reporting them, but they should get fixed too.  Welcome to open source :)

PR #976, opened 2023-12-30
Fix :com {cmd} not recognizes FOREIGN_CMDs

### LINKED ISSUE #973: cmds added via `vifm.cmds.add` are not visible to `:command {name}`
hi, i found that cmds added by `vifm.cmds.add` are not visible to `:command {cmd}`.

to reproduce:
* `vifm.cmds.add({ name = "hello", description = "hello", handler = function() vifm.sb.info("hello!") end })`
* `:command` # the hello command can be seen in the output
* `:command hello` # while this says `No user-defined commands found`

after read some code, i learnt that `vifm.cmds.add` adds `FOREIGN_CMD`, `:command {cmd}` recognizes `USER_CMD` only, and returns the `cmd_t->cmd`. regarding to the output of `:command`, it treats FOREIGN_CMD as user defined commands already, so i think it'd be better to make `:command {cmd}` be aware of FOREIGN_CMD to be more consistent.

if the above support can be added, then what should be showed for `FOREIGN_CMD`s?
* `cmd_t->description` gives less infomation than `cmd_t->cmd`
* `cmd_t->handler` gives no info

haolian9:
Hi, this tries to fix #973.

after this PR, I believe another relevant problem is `:com <tab>`, to be specific, the completion of `:com` should show FOREIGN_CMDs too.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -159,7 +159,8 @@ Completion for fish shell by Hoang Nguyen (a.k.a. FollieHiyuki).
 
 Alexandr Keyp (a.k.a. IAmKapuze) contributed improvements for :compare.
 
-高浩亮 (a.k.a. haolian9) enabled mouse handling.
+高浩亮 (a.k.a. haolian9) enabled mouse handling and fixed displaying of
+foreign commands in `:commands {prefix}`.
 
 Zhipeng Xue fixed a couple of potential NULL dereferences.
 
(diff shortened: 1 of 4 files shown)


xaizek:
Thanks.  And a separate thanks for the test!  I've pushed minor updates and will merge once CI passes (just in case).

xaizek:
I forgot to address:

> after this PR, I believe another relevant problem is `:com <tab>`, to be specific, the completion of `:com` should show FOREIGN_CMDs too.

Yes, that's correct.

PR #977, opened 2023-12-31
Fix :com and :delcom not completing FOREIGN_CMD names

haolian9:


xaizek:
Sorry for the delay and thanks for the changes.  I've moved new test to `tests/cmds` where it belongs and added some related changes along the way, the most important is describing contents of test suites in `HACKING.md` and what sources they mainly target.

PR #986, opened 2024-02-26
improve pack/unpack plugin

dacyberduck:
#### Unpack
 - now supports unpacking multiple files at once. (Windows & BSD requires testing)
 - handles file names with special characters like whitespace. (Windows & BSD requires testing)
 - optionally takes path-to-a-directory as argument and extracts archive in that directory. Eg, `:Unpack /tmp`
 - uses p7zip for common formats like zip, rar, 7z

#### Pack
 - optionally takes archive-name as argument. Eg. `:Pack foobar.zip` `:Pack foo\ \(bar\).tar.gz`
 - merged pack & unpack into one plugin called packer
 
[preview](https://asciinema.org/a/HVKnkHhLYru9arGYguBUg3Wqx)


xaizek:
Thanks, meant to import your changes from [q2a](https://q2a.vifm.info/1500/how-do-i-create-a-filextype-handler-in-lua-plugin), but looks like those pastes aren't even available anymore.

xaizek:
I made some modifications.  Will need to test a bit more and also on Windows before merging.

dacyberduck:
Regarding [1a2cac36](https://github.com/vifm/vifm/pull/986/commits/1a2cac364c4cb952af8562d9b5855f91527f36f2), typical bash doesn't escape characters like '+', '@'. Also ':' is treated somewhat differently. If vifm.escape() doesn't escape those characters then it will probably modify the names.

xaizek:
> Regarding [1a2cac36](https://github.com/vifm/vifm/pull/986/commits/1a2cac364c4cb952af8562d9b5855f91527f36f2), typical bash doesn't escape characters like '+', '@'. Also ':' is treated somewhat differently. If vifm.escape() doesn't escape those characters then it will probably modify the names.

It won't because all slashes that are actually present in the name are doubled. `tar`, however, needs additional escaping of a colon, I've never noticed that it treats `:` specially.

xaizek:
Finally merging this.  Testing on Windows demonstrated some issues with handling of background processes there and `VifmJob:errors()` could return empty string after job is done even if it printed errors on Linux as well.  Had to debug and address that.

PR #992, opened 2024-04-11
Fix handling of external commands starting with `-`

intelfx:
Currently, at least with bash and zsh, vifm is unable to run a command whose name starts with a dash (`-`) inside of a `term()` or `system()` call:

    zsh: bad option string: '-fd-compgen <...>'

Fix this by applying the idiomatic `--` guard to the shell invocation (by convention and by POSIX, `--` denotes end of options).


xaizek:
Thanks.

PR #995, opened 2024-05-02
Fix error and typos in documentation

qadzek:


xaizek:
Thanks. This also indirectly caused CI failure which led to discovering broken formatting, hence a bunch of pushes for debugging.

Issue #1007, opened 2024-07-26
[feature request] adding the `<wait>` map argument to `vifm.keys.add()`


haolian9:
hi folks, i got the need to remap `<c-z>` in a lua plugin through vifm.keys.add(),
as `<c-z>` is builtin, i have to mark it as `<wait>` [i believe](http://q2a.vifm.info/1483/how-to-correctly-display-mapping-completions?show=1483#q1483).

but vifm.keys.add() accepts no wait parameter, so is it possible to add it?

i have read some code and know that: vifm.keys.add() invokes vle_keys_foreign_add while `:map` invokes vle_keys_user_add, so they add two kind of keys. i'm not sure that if a foreign key is supposed to have `wait` property enabled, or maybe the `wait` property should be enabled by default.

haolian9:
it's not easy than i thought, i tried
```
diff --git a/src/engine/keys.c b/src/engine/keys.c
index 621055817..8c0ac546d 100644
--- a/src/engine/keys.c
+++ b/src/engine/keys.c
@@ -1015,6 +1015,8 @@ vle_keys_foreign_add(const wchar_t lhs[], const key_conf_t *info,
                                                          : BUILTIN_WAIT_POINT;
        curr->foreign = 1;
        curr->conf = *info;
+       curr->no_remap = 1;
+       curr->wait = 1;
        return 0;
 }
```
and it makes no difference.

haolian9:
ah, it seems that the `<c-z>` is a special one.

haolian9:
as i know more, now i believe this `<wait>` is irrelevant to my real problem. so i'm closing it as i have no concrete usecase for it so far.

xaizek:
> as i know more, now i believe this `<wait>` is irrelevant to my real problem. so i'm closing it as i have no concrete usecase for it so far.

Yes, it is irrelevant.  `<wait>` effectively raises priority of a user mapping over a builtin one, so your guess that it might play a role was justified if it weren't for special handling.

PR #1009, opened 2024-07-27
make ctrl-z remappable in user land

haolian9:
hi, while trying to figure out a way for #1007, i found that `<c-z>` has been hard-coded in event_loop::get_char_async_loop.

after read the context of get_char_async_loop, i think there is no particular reason to object  on making ctrl-z remappable in user land, so i made this PR.

the real usecase is that i have a plugin which runs vifm in a neovim's terminal buffer, and i want to add a keymap to vifm that informs nvim to hide the terminal buffer. as a proof of concept, i have already made this 'hidding functionality' work with [lhs-<space>.](https://github.com/haolian9/reveal.nvim/blob/master/lua/reveal/vifm/reveal/init.lua#L104).

### DIFF
--- a/src/event_loop.c
+++ b/src/event_loop.c
@@ -210,7 +210,7 @@ event_loop(const int *quit, int manage_marking)
 				}
 			}
 
-			if(c == WC_C_z)
+			if(c == WC_C_z && !vle_keys_user_exists(WK_C_z, vle_mode_get()))
 			{
 				instance_stop();
 				continue;
--- a/tests/misc/event_loop.c
+++ b/tests/misc/event_loop.c
@@ -5,6 +5,7 @@
 #include "../../src/cfg/config.h"
 #include "../../src/engine/cmds.h"
 #include "../../src/engine/keys.h"
+#include "../../src/engine/mode.h"
 #include "../../src/lua/vlua.h"
 #include "../../src/modes/modes.h"
 #include "../../src/modes/wk.h"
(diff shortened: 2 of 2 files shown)


xaizek:
Without this PR: <kbd>Ctrl-Z</kbd> can be pressed in any mode and even in the middle of a shortcut.
With this PR: <kbd>Ctrl-Z</kbd> can be pressed only in normal mode as a separate shortcut.

Handing in `get_char_async_loop()` was probably just an easy way of handling <kbd>Ctrl-Z</kbd> in any mode.  Vim handes it more like in this PR (there is even a special handling in command-line mode which I've never noticed before) but also does it in visual and selection modes.  I think Vifm should handle the key in almost every mode it has.  To not repeat the same handler everywhere, existing code could be modified to look like:
```c
if(c == WC_C_z && !vle_keys_user_exists(WK_C_z, vle_mode_get()))
```
It's not exactly equivalent to adding handlers everywhere though and is a mix of current Vifm's and Vim's behaviours.

haolian9:
ah, thank you! i forgot that other modes can also have this lhs. your approach is much cleaner, and ctrl-z is the only one here that needs special treatment. so i also think there is no need to add identical handler to every mode.

haolian9:
i'm trying to add a test case.

haolian9:
yet it seems `vle_keys_exec` will not go through event_loop.c, which is commonly used in all the tests under `tests/keys/*`. i read the `tests/misc/event_loop.c`, but it helps not.

i feel there is no way to write a test for the change in this PR unfortunately.

xaizek:
`tests/keys` are testing `src/engine/keys.c` alone as mentioned in [HACKING.md](https://github.com/vifm/vifm/blob/master/HACKING.md#package-contents).  `tests/misc/event_loop.c` is the correct place.  You can copy

https://github.com/vifm/vifm/blob/b792d44b3d8a81d03586866a3ab5a49f19965124/tests/misc/event_loop.c#L50-L59

and replace `L"x"` with `WK_C_z`.  That should be enough.

haolian9:
thank you! i did not pay much attention on that before. i have added a test now which differs from your suggestion slightly.

xaizek:
Thanks for the contribution!

PR #1047, opened 2025-01-08
Fix AppStream metadata id and screenshot tag

AsciiWolf:
It should be in valid rDNS format:

https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic

Fixes:

```
$ appstreamcli validate vifm.appdata.xml
W: info.vifm:3: cid-desktopapp-is-not-rdns info.vifm
...
```

Also fix the incorrect "screenshot" tag:

```
I: info.vifm.vifm:16: unknown-tag screenshot
```

### DIFF
--- a/data/vifm.appdata.xml
+++ b/data/vifm.appdata.xml
@@ -13,11 +13,13 @@
 
 	<launchable type="desktop-id">vifm.desktop</launchable>
 
-	<screenshot type="default">
-		<image>
-			https://raw.githubusercontent.com/vifm/vifm/master/data/graphics/screenshot.png
-		</image>
-	</screenshot>
+	<screenshots>
+		<screenshot type="default">
+			<image>
+				https://raw.githubusercontent.com/vifm/vifm/master/data/graphics/screenshot.png
+			</image>
+		</screenshot>
+	</screenshots>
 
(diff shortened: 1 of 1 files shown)


data/vifm.appdata.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop-application">
-	<id>info.vifm</id>
+	<id>info.vifm.vifm</id>
```
xaizek, line 3: `info.vifm.app` would also be a valid id, right?  `info.vifm.vifm` is too weird (`info.vifm` should actually be enough, but looks like spec authors didn't take this common case into account).

AsciiWolf: It would, however "`info.vifm.`" should be the developer/domain name while the last part (`vifm`) should be the app name. On the other hand, there are some apps, for example org.telegram.desktop, that use the same format as you suggested.

xaizek: `org.7-zip.7-zip` is used as an example, so I guess ugly IDs are the norm there.

Thanks for the fixes!

Issue #419, opened 2019-03-27
sixel support


TomJansen:
Hi!
Does Vifm support sixel graphics? It is a relatively robust way to display images. I have tried to make it work in vifm with a sixel compatible terminal but unfortunately, it doesn't work, it just displays text in the preview pane.

In addition to displaying normal sixel files, a nice feature can be to convert an image with img2sixel, and then display it directly on the preview screen of vifm with vifm's filepreview command.

xaizek:
Hello.

It doesn't and I'm not sure it's possible to change this. As far as I know there is no support for it in curses, so it can only work if curses is bypassed. I know there is a function called `putp()`, but I have no idea whether it can help. In examples `putp()` is usually used without curses handling windows on the screen. Using `putp()` and windows can have conflicts.

If there exists a curses application which is capable of displaying sixel graphics, then it must be possible. I'm thinking that it might be worth asking on StackOverflow about it, Thomas Dickey (maintainer of ncurses, xterm, etc.) seems to answer quite a lot of curses-specific questions there and  probably knows the answer.

TomJansen:
Here is an answer from Thomas Dickney regarding sixel implementation in ncurses.
http://lists.gnu.org/archive/html/bug-ncurses/2016-12/msg00017.html

xaizek:
Thanks. Looks like it's not a no, but I'm not sure what "you have to see how to set aside an area on the screen which curses doesn't update" means. If it needs to not contain any curses windows, then it's a no after all. If it means that curses updates will just destroy the image, then it's fine.

I guess I just need to build xterm with sixel support and try it out.

xaizek:
I tried it earlier today and it might be doable.

The image is displayed and almost always stays on the screen (it sometimes disappears on terminal resizing). Unlike with `w3mimg`, where image might disappear too soon, sixel stays on the screen too long and might hover over completely unrelated window (for example on top of `:version`).

Also previous image isn't cleared on moving to the next file.

At least the latter issue needs to be addressed for sixel support to be useful, hovering over other windows seems to be bearable (ideally, this wouldn't happen as well, but it might not be possible to fix).

TomJansen:
>sixel stays on the screen too long and might hover over completely unrelated window (for example on top of :version

I'm just merely brainstorming here, but maybe a hack would be to replace the sixel with an empty sixel image? That is a "?" character, so a full empty sixel file would be `?\`.

xaizek:
[...]

PR #1077, opened 2025-04-24
Add show-sixel script from #419

mcepl:
Cleaned up and fixed to make shellcheck happy.


mcepl:
I spoke too soon. It doesn’t work for me.

PR #1132, opened 2026-02-09
Add .deb handlers examples in the data/vifmrc

krekhovx:
I found an extremely helpful thing for Debian users. I couldn’t figure out for a long time how to replicate the behavior of Midnight Commander, where pressing F3 views detailed package information. I was able to write a filetype handler command that reproduces this behavior in vifm, and I think it would be useful to include it as an example.

### DIFF
--- a/data/vifmrc
+++ b/data/vifmrc
@@ -460,6 +460,10 @@ filextype */
         \ {View in thunar}
         \ Thunar %f &,
 
+" Deb packages
+fileviewer *.deb dpkg-deb -I %c; echo; dpkg-deb -c %c
+filetype *.deb echo "$(dpkg-deb -I %c; echo; dpkg-deb -c %c)" | less
+
 " Syntax highlighting in preview
 "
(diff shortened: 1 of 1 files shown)


data/vifmrc
```diff
@@ -460,6 +460,9 @@ filextype */
         \ {View in thunar}
         \ Thunar %f &,
 
+" Deb packages
+filetype *.deb sh -c '(dpkg-deb -I "%c"; echo; dpkg-deb -c "%c") | less'
```
xaizek: Is the use of `sh -c` a workaround for presence detection?  I think `echo` is better because `%c` is properly escaped for use without quotation and will do well in `$(...)`.  This should work the same:
```vim
filetype *.deb echo "$(dpkg-deb -I %c; echo; dpkg-deb -c %c)" | less
```

krekhovx: I made a rebase, added two examples and tested them, it works well, if something is wrong, let me know

PR #1149, opened 2026-04-04
refactor: run `git status -z` asynchronously to not block directory traversal

stevenxxiu:
I tried to do a todo. Hope this is what you're looking for.

### DIFF
--- a/data/plugins/git/README.md
+++ b/data/plugins/git/README.md
@@ -98,7 +98,6 @@ similar to how `git status --short` does it.
 
 **TODO:**
 
- * Run `git status -z` asynchronously to not block directory traversal.
  * Make cache update faster if directory modification is detected.
  * Check behaviour with submodules.
  * Check if `!!` status can ever be seen.
--- a/data/plugins/git/README.md
+++ b/data/plugins/git/README.md
@@ -99,7 +99,6 @@ similar to how `git status --short` does it.
 **TODO:**
 
  * Make cache update faster if directory modification is detected.
- * Check behaviour with submodules.
  * Check if `!!` status can ever be seen.
  * Consider using `git status --porcelain=v2 -z` as it provides more
    information.
--- a/data/plugins/git/README.md
+++ b/data/plugins/git/README.md
@@ -57,7 +57,7 @@ similar to how `git status --short` does it.
 
 | Value | Meaning
 | ----- | -------
-| `  `  | Ignored or unchanged.
+| `  `  | Ignored.
 | ` M`  | Modified in worktree.
 | `MM`  | Modified in index and in worktree.
 | `AM`  | Added in index, modified in worktree.
@@ -74,6 +74,7 @@ similar to how `git status --short` does it.
 | `DU`  | Deleted here, but updated in merged changes.
 | `AA`  | Added here and in merged changes.
 | `UU`  | Modified here and in merged changes.
+| `GG`  | Unchanged.
 | `??`  | Untracked.
 
 **Possible column values for directories:**
@@ -93,13 +94,16 @@ similar to how `git status --short` does it.
 
 **Issues:**
 
- * Has a noticeable initial delay for large Git repositories due to fetching
-   statuses synchronously.
+ * Large repositories won't show complete status due to hitting pipe size limit.
+ * Git status for a submodule isn't fetched immediately after entering it.
 
 **TODO:**
 
- * Address obtaining GG entries failing for large repositories due to filling
-   of the pipe before `onexit` callback in invoked.
+ * Could add `:Gopt` to make some behaviour optional:
+   - asynchronous/synchronous: as a workaround for large repositories
+   - marking unchanged file: also workaround but also a user's choice (e.g.,
+                             marking unchanged or ignored files)
+   - discovering git repositories and showing their statuses
  * Reuse old cache for entire subtree while it's being updated (only a specific
    path gets to reuse it right now).
  * Make cache update faster if directory modification is detected.
--- a/data/plugins/git/README.md
+++ b/data/plugins/git/README.md
@@ -98,6 +98,8 @@ similar to how `git status --short` does it.
 
 **TODO:**
 
+ * Address obtaining GG entries failing for large repositories due to filling
+   of the pipe before `onexit` callback in invoked.
  * Reuse old cache for entire subtree while it's being updated (only a specific
    path gets to reuse it right now).
  * Make cache update faster if directory modification is detected.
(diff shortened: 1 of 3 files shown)


stevenxxiu:
Is there a way to redraw the *Git* column, after `vifm.startjob()` and updating info? My current hack of `view:cd()` isn't good, as it collapses directories.

xaizek:
> I tried to do a todo. Hope this is what you're looking for.

Nice, thanks.  I was imagining a more complicated implementation with a queue, but I guess cache nodes act similarly to a queue at least for an individual path.

> Is there a way to redraw the _Git_ column, after `vifm.startjob()` and updating info? My current hack of `view:cd()` isn't good, as it collapses directories.

The view needs to be redrawn, but there isn't really a way (there needs to be one).  The most optimal method right now is probably:
```lua
vifm.opts.global.laststatus = not vifm.opts.global.laststatus
vifm.opts.global.laststatus = not vifm.opts.global.laststatus
```

stevenxxiu:
I marked committed and unmodified files with `GG`. Marking those is what `git-nautilus-icons` and what past file manager *Git* software I used did. Maybe this means ignored files don't need to be marked separately?

stevenxxiu:
How do I use `libgit2` here? Does something need to change on the *C* side?

xaizek:
> How do I use `libgit2` here?

I think https://github.com/q66/cffi-lua should allow doing that.

> Does something need to change on the _C_ side?

If changes are necessary to make `cffi-lua` work (e.g., if `require()` has issues), then yes.

stevenxxiu:
Am I right in saying the build process needs to be updated to include `cffi-lua` here then?

xaizek:
> Am I right in saying the build process needs to be updated to include `cffi-lua` here then?

Sorry, missed this comment.  No, I was thinking of it being an external dependency of a plugin, installed by some package manager (I think `luarocks` is mostly used for this).  But enabling such interoperation is what may need changes in the C code of Vifm (maybe not and search path could be updated in Lua at least as a workaround).

***

Thanks for the updates, I've added some fixes and will merge the PR.

One drawback of reading all output of asynchronous processes is that large output cannot be obtained because a filled up pipe blocks the producer, so `onexit` never gets to run.  This is mostly relevant for `GG` and medium to large repositories because `git ls-tree` can produce much larger output than `git status`.  It can't be addressed nicely at this point (need a data availability callback or something similar), although a redirection to a temporary file can be a workaround.

stevenxxiu:
I tested it out and this commit <https://github.com/vifm/vifm/pull/1149/changes/cd7cda5c94be6e0e8cec2c83d81929390cbb5036> means that committed files no longer show up as having `GG`. I use this to track which files are in the repo and which files are ignored. This is what `git-nautilus-icons` and what past file manager *Git* software I used did. Can committed files be shown?

(Actually I realized my own branch was a bit buggy here. Going in a folder with a repo and out marks that folder as being a *Git* repo too.)

Thanks for merging this.

xaizek:
That commit merely fixed a broken check that's used to determine if an optimization is possible.  Compatibility of the optimization with `GG` status is a separate question and I suppose these are in contradiction.  However, doing a synchronous check to avoid running `git status` asynchronously doesn't really much sense, so I've nuked it.  I've also added `:Gopt unchanged/nounchanged` to make `GG` optional (on by default; I just find it noisy to mark every file).

stevenxxiu:
@xaizek Thanks. This has the same bug though. Go in a directory containing a repo that has an untracked file. Then the status correctly shows up. But then go up to the parent directory. This directory that's unrelated to *Git* shows that it has an untracked file too.

stevenxxiu:
I tried to debug this with *MobDebug* installed from *LuaRocks*, but I get the error:

```
Failed to start 'git' plugin: error loading module 'socket.core' from file '/home/steven/.config/luarocks/lib/lua/5.5/socket/core.so':
        dynamic libraries not enabled; check your Lua installation
```

I think this means the build process needs to change? This also ties into using *LuaRocks* for `cffi-lua`.

xaizek:
> This has the same bug though. Go in a directory containing a repo that has an untracked file. Then the status correctly shows up. But then go up to the parent directory. This directory that's unrelated to _Git_ shows that it has an untracked file too.

I see it now.  Tried to reproduce before, but somehow missed it.

> I think this means the build process needs to change?

Yeah, looks like `#define LUA_USE_DLOPEN ` is missing.  Lua's configuration file detects Windows but not anything else, so dynamic libraries for Lua are enabled only on Windows right now.

stevenxxiu:
I managed to start a debugger, but it doesn't break. Do you know how I can debug *Lua* plugins? [lua_dynamic_library_loading.patch](https://github.com/user-attachments/files/26721321/lua_dynamic_library_loading.patch).

    $ luarocks --local --lua-version=5.4 install mobdebug

Then I downloaded and configured <https://github.com/atomicptr/mobdap/releases/tag/v0.2.1> for *Neovim*.

xaizek:
> Do you know how I can debug _Lua_ plugins?

No, I was using `print`-debugging for complicated cases so far (<kbd>e</kbd> in `:plugins` shows those prints or just do `print = vifm.sb.info` and then `:messages`).  Lua wasn't selected because I'm all that familiar with it but because it was meant for this kind of uses and performs decently.  Will need to look into debugging it more though, thanks for taking first steps in that direction.

This seems to fix it:
```diff
data/plugins/git/statuses.lua | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git a/data/plugins/git/statuses.lua b/data/plugins/git/statuses.lua
index eb11a4803..c48dc6957 100644
--- a/data/plugins/git/statuses.lua
+++ b/data/plugins/git/statuses.lua
@@ -79,7 +79,9 @@ local function set_file_status(node, path, status, expires)
         -- keep the first status
         if node.items[path] == nil then
             node.items[path] = status
-            update_dir_status(node, status)
+            if node.in_git then
+                update_dir_status(node, status)
+            end
         end
         return node.status
     end
```
Status of a file in a repository influences status of its parent directory and repositories are doing the same to their parent.  Actually, this is likely a better fix:
```diff
data/plugins/git/statuses.lua | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/data/plugins/git/statuses.lua b/data/plugins/git/statuses.lua
index eb11a4803..b29c9c097 100644
--- a/data/plugins/git/statuses.lua
+++ b/data/plugins/git/statuses.lua
@@ -120,7 +120,7 @@ function update_subdir(sub_at, path, node)
         onexit = function(sub_job)
             local sub_status_all = sub_job:stdout():read('a')
             local status = sub_status_all == '' and 'GG' or sub_status_all:sub(1, 2)
-            set_file_status(node, path, status, node.expires)
+            node.items[path] = status
             redraw()
         end
     }
```

PR #1151, opened 2026-04-15
Some fixes to the *Git* plugin

stevenxxiu:
I included your <https://github.com/vifm/vifm/pull/1149#issuecomment-4247455941> too under your name.


xaizek:
Thanks for confirming that was the fix and providing another one.

PR #1153, opened 2026-04-24
refactor(git): switch to `libgit2`

stevenxxiu:
This requires:

- `libgit2` being installed on the OS.
- `luarocks --local --lua-version=5.4 install cffi-lua`.

There's still a small issue where when I go in some *Git* repos, the status of children don't show up. Haven't figured this out yet.


stevenxxiu:
Another issue is when I go in and out of `vifm/autom4te.cache/` a few times, after perhaps the cache expires, it turns from nothing into something else.

I might need some help on these 2 issues.

xaizek:
I've added a couple fixes and identified an issue with filling the cache, although not sure that it's responsible for both problems (in particular I'm puzzled why `autom4te.cache` has no status).  [This line](https://github.com/stevenxxiu/vifm/blob/944e0dee95d21294d7914282231646aa0c7427f4/data/plugins/git/statuses.lua#L356) uses node for the current path (`at`) but `update_status()` obtains and fills status relative to the repository's root (`root`).

PR #1154, opened 2026-04-30
Add 'keepsel' option for a more persistent selection

cairo55:
vifm often inconveniently resets the selection, for example after cancelling command-line entry. I understand that this is made to mimic how vim behaves, but I find it to be unhelpful in a file manager.  

This pull request adds an option `'keepsel'`, that when set makes it so vifm behaves a little more intuitively, and closer to how other terminal file managers do (ranger, lf, mc). It is disabled by default.

Thank you.

### DIFF
--- a/src/modes/cmdline.c
+++ b/src/modes/cmdline.c
@@ -1719,7 +1719,7 @@ cmd_return(key_info_t key_info, keys_info_t *keys_info)
 	if(prev_mode == VISUAL_MODE && sub_mode != CLS_VFSEARCH &&
 			sub_mode != CLS_VBSEARCH)
 	{
-		modvis_leave(curr_stats.save_msg, !cfg.keepsel, !cfg.keepsel);
+		modvis_leave(curr_stats.save_msg, !cfg.keepsel, 0);
 	}
 
 	if(sub_mode == CLS_COMMAND || sub_mode == CLS_MENU_COMMAND)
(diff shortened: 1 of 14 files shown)


cairo55:
It seems that `flist_sel_stash_if_nonempty` is redundant, `save_selection` already terminates early if nothing is selected.

src/modes/cmdline.c
```diff
@@ -1723,7 +1719,7 @@ cmd_return(key_info_t key_info, keys_info_t *keys_info)
 	if(prev_mode == VISUAL_MODE && sub_mode != CLS_VFSEARCH &&
 			sub_mode != CLS_VBSEARCH)
 	{
-		modvis_leave(curr_stats.save_msg, 1, 0);
+		modvis_leave(curr_stats.save_msg, !cfg.keepsel, !cfg.keepsel);
```
xaizek:
```suggestion
		modvis_leave(curr_stats.save_msg, !cfg.keepsel, 0);
```
The command-line command is executed below and clearing selection here makes it impossible to run commands on visually selected files.

xaizek (changes requested):
Thank you for the patches!

> closer to how other terminal file managers do (ranger, lf, mc)

This explains why some users struggle with selection in Vifm.  After using it almost exclusively for 15 years it becomes hard to compare the behaviour to other file managers.

> It seems that `flist_sel_stash_if_nonempty` is redundant, `save_selection` already terminates early if nothing is selected.

Good point.  It must have been a small optimization, but I see how to do it directly in `flist_sel_stash()` instead.

cairo55:
@xaizek Thanks, and thanks for the work on vifm, it's great.

PR #1158, opened 2026-05-16
Windows: fix sixel/pass-through preview via CONOUT$ and VT mode

### LINKED ISSUE #1157: Windows: fix sixel/pass-through preview (PDCurses strips ENABLE_PROCESSED_OUTPUT)
## Problem

On Windows, fileviewers that use the `%pd` macro (`VK_PASS_THROUGH`) produce no visible output in the preview pane — raw escape sequences appear as literal characters, or output goes to the wrong buffer entirely.

Three issues combine to cause this:

### 1. Wrong screen buffer

PDCurses calls `SetConsoleActiveScreenBuffer()` during `initscr()`, creating a new active buffer. `GetStdHandle(STD_OUTPUT_HANDLE)` continues to return the original, now-inactive buffer — writes there are invisible. The fix is to open `\\.\CONOUT$`, which always refers to the currently active screen buffer.

### 2. Console mode strips ESC processing

PDCurses sets the output mode to `ENABLE_LVB_GRID_WORLDWIDE` (0x10) only, removing `ENABLE_PROCESSED_OUTPUT` (0x01). Without `ENABLE_PROCESSED_OUTPUT`, ESC bytes are never treated as control character starts, so `ENABLE_VIRTUAL_TERMINAL_PROCESSING` (0x04) has no effect — the VT engine never fires. Sixel DCS sequences appear verbatim.

The fix sets `ENABLE_PROCESSED_OUTPUT | ENABLE_VIRTUAL_TERMINAL_PROCESSING` before writing VT sequences, then restores the original mode (0x10) immediately after so PDCurses' next redraw cycle is unaffected.

### 3. Sixel pixels survive character-cell writes

In Windows Terminal, sixel graphics occupy a separate rendering layer. PDCurses' normal redraw via the Console API does not erase them. When the user exits preview mode, sixel remnants remain visible behind the file listing. Erasing requires writing spaces via the VT path with an explicit SGR reset (`\x1b[0m`).

## Environment

- Windows Terminal with sixel support
- PDCurses (win32/wincon backend)
- Viewer: `chafa -f sixels ... %pd`
- Confirmed on vifm 0.15, MinGW (`x86_64-w64-mingw32-gcc`)

## Fix summary

- `ui_pass_through()` — open `CONOUT$`; enable VT mode around `WriteFile`; restore mode after
- `ui_pass_through_clear()` — new function; fills window cells with SGR-reset spaces via the same VT path to erase sixel pixels when switching away from a pass-through preview
- `refresh_view_win()` — skip PDCurses redraw while pass-through graphics are on screen; call `ui_pass_through_clear()` when preview is toggled off
- `wipe_area()` / `qv_draw()` — call `ui_pass_through_clear()` before PDCurses-based clearing on file cursor move and preview close
- `get_lines()` — use `VC_SYNC` for `VK_PASS_THROUGH` viewers so data is available on the first render rather than requiring an async redraw cycle

A patch and pull request follow.

Maxim4711:
Fixes #1157

## What

Makes `%pd` (`VK_PASS_THROUGH`) fileviewers work correctly on Windows — specifically sixel image viewers such as `chafa -f sixels`.

## Root cause

Three separate issues on Windows prevent pass-through output from rendering:

**1. Wrong screen buffer.**
PDCurses calls `SetConsoleActiveScreenBuffer()` during `initscr()`, so `GetStdHandle(STD_OUTPUT_HANDLE)` returns the original, now-inactive buffer. Writes there are invisible. Fix: open `\\.\CONOUT$`, which always resolves to the currently active screen buffer.

**2. Console mode strips ESC processing.**
PDCurses sets the output mode to `ENABLE_LVB_GRID_WORLDWIDE` (0x10) only, which removes `ENABLE_PROCESSED_OUTPUT` (0x01). Without it, ESC bytes are not treated as control-sequence starts, so `ENABLE_VIRTUAL_TERMINAL_PROCESSING` has no effect and sixel DCS sequences appear as literal text. Fix: set both flags before writing, restore the original mode immediately after so PDCurses' next redraw is unaffected.

**3. Sixel pixels survive character-cell writes.**
Windows Terminal renders sixel graphics in a separate layer that PDCurses' Console API writes do not erase. When the user closes the preview pane, sixel remnants remain visible behind the file listing. Fix: new `ui_pass_through_clear()` covers every cell of a window with an SGR-reset space (`\x1b[0m `) via the same VT path.

## Changes

| File | Change |
|---|---|
| `src/ui/ui.h` | Declare `ui_pass_through_clear()` |
| `src/ui/ui.c` | `ui_pass_through()`: use `CONOUT$` + VT mode for Windows |
| `src/ui/ui.c` | New `ui_pass_through_clear()`: erase sixel pixels via VT path |
| `src/ui/ui.c` | `refresh_view_win()`: skip PDCurses redraw while graphics are shown; call clear on preview toggle-off |
| `src/ui/quickview.c` | `wipe_area()`: call clear before PDCurses fill when kind != VK_TEXTUAL |
| `src/ui/quickview.c` | `qv_draw()`: call clear before `ui_view_erase()` on file cursor move |
| `src/ui/quickview.c` | `get_lines()`: use `VC_SYNC` for `VK_PASS_THROUGH` so data is ready on first render |

All Windows-specific code is inside `#ifdef _WIN32`. No behaviour change on other platforms.

## Testing

Tested on Windows 11 with Windows Terminal (sixel enabled) and the following vifmrc viewer:

```vim
fileviewer {*.jpg,*.jpeg,*.png,...},<image/*>
    \ chafa -f sixels --passthrough=none --font-ratio=0.5 -s %pwx%ph --animate=false --margin-bottom=0 %"c:p %pd
```

- Image renders correctly in the preview pane on cursor move
- Sixel pixels are fully erased when closing the preview pane (`w`)
- Sixel pixels are fully erased when switching to a non-image file while preview is open

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -201,3 +201,5 @@ plugin.
 cairo55 implemented 'keepsel' option.
 
 Vladimir Baranov exposed `os.remove()` in Lua.
+
+Maxim4711 provided a patch to make dispalaying Sixel on Windows Terminal work.
--- a/BUGS
+++ b/BUGS
@@ -24,3 +24,4 @@
   non-POSIX shells which have `-c` option accept a value.
 * Incrementing/decrementing file names doesn't work as expected for numbers
   outside [-2**63; -2**63) range.
+* Sixel doesn't get cleared in Windows Terminal on moving panes around.
--- a/ChangeLog
+++ b/ChangeLog
@@ -58,6 +58,10 @@
 	expression but is used to specify patterns like `{*.pdf}!/abc/` after the
 	plus sign.  Thanks to filterfalse, tagwint and CaptainFantastic.
 
+	Added somewhat incomplete handling of Sixel for Windows Terminal.
+	Clearing doesn't happen if moving panes while Sixel is displayed, but it
+	works otherwise.  Patch by Maxim4711.
+
 	Updated utf8proc to v2.11.3.
 
 	Made documentation on which :commands can have comments a bit more
--- a/src/ui/quickview.c
+++ b/src/ui/quickview.c
@@ -205,7 +205,17 @@ qv_draw(view_t *view)
 		return;
 	}
 
+	/* Clear any sixel graphics from the previous frame before erasing the
+	 * window and drawing new content.  PDCurses character cells alone do not
+	 * cover sixel pixels in Windows Terminal. */
+	if(other_view->displays_graphics)
+	{
+		ui_pass_through_clear(other_view->win);
+	}
+
 	ui_view_erase(other_view, 1);
+
+	other_view->displays_graphics = 0;
 
 	curr = get_current_entry(view);
 	if(!fentry_is_fake(curr))
@@ -220,6 +230,7 @@ qv_draw(view_t *view)
 			.h = ui_qv_height(other_view),
 		};
 		(void)view_entry(curr, &parea, &qv_cache);
+		other_view->displays_graphics = (qv_cache.kind != VK_TEXTUAL);
 	}
 
 	refresh_view_win(other_view);
@@ -358,6 +369,7 @@ view_file(const char path[], const preview_area_t *parea,
 
 	update_cache(cache, path, viewer, kind, parea, max_lines);
 	strlist_t lines = get_lines(cache);
+
 	draw_lines(&lines, cfg.wrap_quick_view, &cache->pa, cache->kind);
 
 	if(cache->kind != VK_TEXTUAL)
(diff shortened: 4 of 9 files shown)


xaizek:
Thanks, this does seem to work and explains why a few of my attempts didn't.  However, some implementation choices are rather dubious.  I'll need to see about not disabling caching, not wiping screen unnecessarily and whether patching PDCurses would make things look better.

Maxim4711:
Surely, the suggested patch was to large extent written using ai, so - i am convinced, you can improve it dramatically - i can only confirm - what i tested - image preview works properlyx using this build, If you need some more data from my environment (like vifmrc or - some more details about windows build/version - i'll provide it.

xaizek:
Finally got around to testing this more.  Updated changes to work with `'millerview'` better.  Also found a few issues with `'millerview'` not caused by these changes, those will be pushed to `master` directly.  Pushed this branch to have GitHub close the PR as merged, now waiting for CI just in case something will be off.

> [...] whether patching PDCurses would make things look better.

PDCurses sets console mode to hard-coded values in various places, so nothing was moved there.

***

EDIT: there is no clearing of Sixel when a pane that contains it is moved, but this is left for the future as I want to release v0.15 and changes required to address this problem can break something else.

PR #1165, opened 2026-09-07
Refer to 'millerview' instead of 'miller' in sample vifmrc

ccjmne:
Hi, I just started using vifm and am having a blast so far. Thank you very much!

There's a small error in the comment to the newly added `milleroptions` in the sample vifmrc: `set miller` won't help the user, `set millerview` will.
This modest patch rectifies that.

I see that you keep your commit messages generally under 50, and this one is just 51 characters long: I can probably modify the subject somewhat if it's preferable :-)

### DIFF
--- a/data/vifmrc
+++ b/data/vifmrc
@@ -143,8 +143,8 @@ endif
 " Add -s to the default value to suppress "Permission denied" errors
 set grepprg="grep -n -H -I -r -s %i %a %s"
 
-" Enable preview of files in the right miller column (add `set miller` to turn
-" it on)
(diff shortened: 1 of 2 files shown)


xaizek:
> Hi, I just started using vifm and am having a blast so far. Thank you very much!

Hi, glad to hear that.  Thanks for fixing the comments.
 
> I see that you keep your commit messages generally under 50, and this one is just 51 characters long: I can probably modify the subject somewhat if it's preferable :-)

That extra character isn't worth the trouble :-)

